PR #13, opened 2023-04-25
Android Implementation

nplasterer:
This adds the android side of the native module

https://user-images.githubusercontent.com/3522670/234422981-56b6f29f-b579-4404-bfce-ed7eb0224984.mp4

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -10,16 +10,18 @@ import org.xmtp.android.library.Client
 import org.xmtp.android.library.Conversation
 import org.xmtp.android.library.SigningKey
 import org.xmtp.android.library.XMTPException
+import org.xmtp.android.library.messages.InvitationV1ContextBuilder
 import org.xmtp.android.library.messages.PrivateKeyBuilder
 import org.xmtp.android.library.messages.Signature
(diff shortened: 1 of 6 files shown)


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -1,47 +1,266 @@
...
+            }
+            val conversation =
+                findConversation(topic = conversationTopic, conversationId = conversationID)
+                    ?: throw XMTPException("no conversation found for $conversationTopic")
+            val preparedMessage = conversation.prepareMessage(content = content)
+            val decodedMessage = preparedMessage.decodedMessage()
+            preparedMessage.send()
+            DecodedMessageWrapper.encode(decodedMessage)
+        }
+        // TODO: Support conversationId
```
elisealix22, line 150: It looks like it handles conversationId -- is there something else we need to support?

elisealix22 (approved):

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
nplasterer: ahh these comments were just copied over from the iOS side @nakajima Are these todos old? Should I remove them from both

nakajima: Yea the other thing we need to support is just me cleaning up comments I've left to myself. We can remove these from both.

PR #20, opened 2023-04-27
Docs: Add npm command to README

jhaaaa:


nplasterer (approved):
This seems fine for now. The one thing on android side is the minimum sdk needs to be `minSdkVersion = 22` to work with our SDK

PR #31, opened 2023-05-16
Fix build on iOS

nakajima:
We were using an outdated version of the RN push notification library for iOS. This PR updates to use the most recent version.


example/ios/Podfile
```diff
@@ -49,6 +49,7 @@ target 'xmtpreactnativesdkexample' do
   flags = get_default_flags()
 
   pod 'secp256k1.swift', :modular_headers => true
+  pod 'RNCPushNotificationIOS', :path => '../node_modules/@react-native-community/push-notification-ios'
```
michaelx11: this looks brittle - the pod doesn't get pulled in normally? i.e. via expo hooks checking npm packages

michaelx11 (approved):
LGTM - one question around the Podfile path-based item

example/ios/Podfile
nakajima: It was something I pulled in from the docs here: https://reactnative.dev/docs/pushnotificationios.html, after removing it stuff still worked though so I don't think we need it. Removed in https://github.com/xmtp/xmtp-react-native/pull/31/commits/c97c0e93175ae126833d67cdb88eb467106a9900.

nakajima: Then actually removed in https://github.com/xmtp/xmtp-react-native/pull/31/commits/655762c2408345f07cc98364b21f4972fd49d309.

nplasterer (approved):
Sorry about breaking that!

daria-github (approved):
Amazing that this did it, thanks for finding the fix.

PR #27, opened 2023-05-04
Push notifications

nplasterer:
Android part of https://github.com/xmtp/xmtp-react-native/issues/22

https://user-images.githubusercontent.com/3522670/236349849-042ff434-49dd-45b8-a57b-b538575a072a.mp4

Will do iOS in a follow up.

PR #32, opened 2023-05-17
Remove push from example app

nplasterer:
When adding support for push notifications here: https://github.com/xmtp/xmtp-react-native/pull/27. I also added support for push notifications in the example app. This however is a bit complicated and caused build issues.

Instead I'm going to pull the push example out into a separate branch for referencing.


example/ios/Podfile.lock
```diff
@@ -667,6 +661,6 @@ SPEC CHECKSUMS:
   XMTPRust: 2c356f57f034d771717933a6127a7d51667f9843
   Yoga: ba09b6b11e6139e3df8229238aa794205ca6a02a
 
-PODFILE CHECKSUM: c863de1064bef6a085e5c1c74f7ce05cd040466a
+PODFILE CHECKSUM: 522d88edc2d5fac4825e60a121c24abc18983367
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.11.3
```
michaelx11, line 666: No change needed, but I'm on 1.12.0 and Pat is using 1.12.1 (latest) IIRC. There is a chance that Cocoapod version will matter at some point but shouldn't be an issue now.

michaelx11 (approved):

daria-github (approved):

example/ios/Podfile.lock
nplasterer: O this is interesting. I'll get my cocoapods updated.

PR #37, opened 2023-05-18
Update README.md

nplasterer:
We are adding lots of new features so removing the list so that it doesn't confuse people looking at the repo.


neekolas (approved):

jhaaaa (approved):
Thank you for this @nplasterer!

PR #47, opened 2023-06-01
feat: (iOS) use batchQuery instead of one-at-a-time for messages

dmccartney:
This replaces the placeholder implementation of batch-message-listing on iOS with an actual call to batchQuery.

This depends on landing and using https://github.com/xmtp/xmtp-ios/pull/106


nplasterer (approved):
As long as the unit tests still pass looks good to me!

PR #56, opened 2023-06-13
Initial setup work for content types

daria-github:
Added initial classes, types and tests to support registering, encoding, and decoding a codec. 

Screenshot 2023-06-14 at 11 50 18 AM


example/src/tests.ts
```diff
@@ -63,3 +65,55 @@ test("canMessage", async () => {
...
+  registry.register(numberCodec);
+
+  try {
+    const id = numberCodec.contentType.id();
+    const codec = registry.find(id);
+
+    const encodedContent = codec.encode(3.14);
+    const invalidContentToDecode = {
+      ...encodedContent,
+      // @ts-ignore
```
nakajima: What are we ts-ignoring here?

example/src/test_utils.ts
```diff
@@ -0,0 +1,43 @@
...
+      versionMajor: 1,
+      versionMinor: 1,
+    });
+  }
+
+  encode<T>(content: T) {
+    const encodedContent = {
+      type: this.contentType,
+      content: JSON.stringify(content),
+      compression: Compression.COMPRESSION_DEFLATE,
```
nakajima: I think we might want to leave this out because we're not actually doing any compression at the moment.

src/XMTP.types.ts
```diff
@@ -1,7 +1,28 @@
...
+ * Represents encoded content and its metadata.
+ *
+ * @param {ContentTypeID | {}} type - The content type ID or object for this content.
+ * @param {string} content - The encoded content data.
+ * @param {Compression} compression - The compression algorithm used, if any.
+ * @param {string} fallback - A fallback representation of the content, if any.
+ */
+
+export type EncodedContent = {
+  type: ContentTypeID | {};
```
nakajima: When would this just be an empty object?

src/XMTP.types.ts
```diff
@@ -1,7 +1,28 @@
...
+  content: string;
+  compression?: Compression;
```
nakajima: Can this be optional?

src/XMTP.types.ts
```diff
@@ -1,7 +1,28 @@
...
+ *
+ * @param {ContentTypeID | {}} type - The content type ID or object for this content.
+ * @param {string} content - The encoded content data.
+ * @param {Compression} compression - The compression algorithm used, if any.
+ * @param {string} fallback - A fallback representation of the content, if any.
+ */
+
+export type EncodedContent = {
+  type: ContentTypeID | {};
+  content: string;
```
nakajima: I think content here wants to be a `UintArray8` instead of a string since we'll want to allow binary data here.

example/src/tests.ts
daria-github: This `ts-ignore` is needed because this test is specifically testing the error state for an invalid codec, which clashes with the types we expect here.

example/src/test_utils.ts
daria-github: Alright, logic here was to add in what we're going to need to consume in the next few weeks, but will remove so we don't introduce unused code.

src/XMTP.types.ts
daria-github: Removing anyways per earlier comment.

src/XMTP.types.ts
daria-github: Good catch, it would have been a possible state before I removed the `EncodedContent` class but I missed removing it here. Just updated that. Thanks!

src/XMTP.types.ts
daria-github: Will make the update - thanks!

src/XMTP.types.ts
```diff
@@ -1,7 +1,24 @@
...
+ *
+ * Represents encoded content and its metadata.
+ *
+ * @param {ContentTypeID} type - The content type ID for this content.
+ * @param {string} content - The encoded content data.
+ * @param {string} fallback - A fallback representation of the content, if any.
+ */
+
+export type EncodedContent = {
+  type: ContentTypeID;
```
nakajima, line 21: We'll still want to add the [`parameters`](https://github.com/xmtp/proto/blob/main/proto/message_contents/content.proto#L35) field from the EncodedContent protobuf definition since it's central to a few types of content types.
```suggestion
  type: ContentTypeID;
  parameters: { [key: string]: [value: string] }? ,
```

src/XMTP.types.ts
```diff
@@ -1,7 +1,24 @@
...
 export type XMTPViewProps = {
   name: string;
 };
+
+/*
+ *
+ * Represents encoded content and its metadata.
+ *
+ * @param {ContentTypeID} type - The content type ID for this content.
+ * @param {string} content - The encoded content data.
```
nakajima:
```suggestion
 * @param {Uint8Array} content - The encoded content data.
```

nakajima (approved):
Left a couple small things but after those, looks good to me!

PR #60, opened 2023-06-22
Stream all messages

### LINKED ISSUE #9: Support streamAllMessages()
https://github.com/xmtp/xmtp-js#listen-for-new-messages-in-all-conversations

nplasterer:
Closes: https://github.com/xmtp/xmtp-react-native/issues/9

This adds the ability to stream all messages

### DIFF
--- a/src/lib/Conversations.ts
+++ b/src/lib/Conversations.ts
@@ -1,5 +1,6 @@
 import { Client } from "./Client";
 import { Conversation } from "./Conversation";
+import { DecodedMessage } from "./DecodedMessage";
 import * as XMTPModule from "../index";
 
 type Context = {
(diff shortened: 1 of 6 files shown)


src/lib/DecodedMessage.ts
```diff
@@ -1,8 +1,20 @@
-export type DecodedMessage = {
+export class DecodedMessage {
```
nakajima: Why did we need to make this a class?

nakajima (approved):

src/lib/DecodedMessage.ts
nplasterer: @nakajima Maybe this is just my lack of knowledge of typescript but I couldn't figure out how to make a DecodeMessage value on this line https://github.com/xmtp/xmtp-react-native/pull/60/files#diff-c5227210ea741b70a7f1c730d3dc81ff6c90156a21cee65f6fbb29f86099c9f8R67. I kept getting

```
'DecodedMessage' only refers to a type, but is being used as a value here.
```

src/lib/Conversations.ts
```diff
@@ -50,4 +51,21 @@ export default class Conversations {
...
+    XMTPModule.subscribeToAllMessages(this.client.address);
+    XMTPModule.emitter.addListener(
+      "message",
+      async (message: DecodedMessage) => {
+        if (this.known[message.id]) {
+          return;
+        }
+
+        this.known[message.id] = true;
+        await callback(new DecodedMessage(message));
```
nakajima: I think this can just be:
```suggestion
        await callback(message as DecodedMessage);
```

PR #63, opened 2023-06-23
Send/Receive Encoded Content

daria-github:
This PR adds the ability to send and receive encoded content on iOS and Android.
Android -
Screenshot 2023-06-29 at 5 52 24 PM

iOS -


nplasterer:
I don't think you can do the kotlin modules in the next PR. It needs to be done in this one. This is a breaking change. Sending messages won't work at all in android now because you removed the `sendMessage` function. Since other people are cutting npm releases in this repo you can't guarantee that a release won't get cut with this change making the android experience entirely broken.

daria-github:
Makes sense, will do this in 1 go — just updated the overview. Thanks!

example/src/ConversationView.tsx
```diff
@@ -8,25 +9,44 @@ import {
...
+import { TextCodec } from "./test_utils";
+import {
+  CodecRegistry,
+  ContentCodecInterface,
+} from "../../src/lib/CodecRegistry";
 
 type Props = NativeStackScreenProps<RootStackParamList, "Conversation View">;
 
+type DecodedMessage = {
+  content: string;
```
nakajima: I think this should be `any` now that content can be whatever there's a content codec for?
```suggestion
  content: any;
```

example/src/test_utils.ts
```diff
@@ -1,41 +1,85 @@
...
       versionMinor: 1,
     });
   }
 
-  encode(content: Uint8Array) {
+  encode(content: string) {
     const encodedContent = {
       type: this.contentType,
       content: Buffer.from(JSON.stringify(content)),
       fallback: "fallbackText",
```
nakajima: I think the fallback for text should just be the content string itself?

ios/Wrappers/BinaryDataWrapper.swift
```diff
@@ -0,0 +1,24 @@
...
+enum BinaryDataWrapperError: Swift.Error {
+	case encodeError(String)
+}
+
+protocol BinaryDataWrapper: Codable {
+	associatedtype T
+	static func wrap(model: T) throws -> Self
+}
+
+extension BinaryDataWrapper {
```
nakajima, line 18: Not an issue for this PR, but I'm curious how this performs vs the JSON wrapper we're using for the other types. If it's faster we might want to move the other wrappers to use this as well. Just an idle thought, no action needed here.

ios/Wrappers/BinaryDataWrapper.swift
```diff
@@ -0,0 +1,24 @@
...
+  static func encode(_ model: T) throws -> [UInt8] {
+    let wrapper = try wrap(model: model)
+	let encodedData = try MessagePackEncoder().encode(wrapper)
```
nakajima, line 21: Looks like we've got some tabs vs spaces issues in this file.

ios/Wrappers/EncodedMessageWrapper.swift
```diff
@@ -0,0 +1,24 @@
...
+import Foundation
+import XMTP
+
+// Wrapper around XMTP.EncodedMessage to allow passing these objects back
+// into react native.
+struct EncodedMessageWrapper: BinaryDataWrapper {
+	static func wrap(model: XMTP.DecodedMessage) throws -> EncodedMessageWrapper {
+		return EncodedMessageWrapper(
+			id: model.id,
+            content: [UInt8](try model.encodedContent.serializedData()),
```
nakajima: Looks like more tabs vs spaces issues.

src/index.ts
```diff
@@ -58,18 +64,26 @@ export async function listMessages(
...
+    return decode(message);
   });
+
+  for (const message of mappedMessages) {
+    const encodedContent = proto.content.EncodedContent.decode(message.content);
+
+    message.content = encodedContent;
+  }
+
+  return mappedMessages;
```
nakajima: Maybe we should move this into the `map` above so we don't need to loop over the message list twice?

nakajima (approved):
Left a couple small questions/comments, I think we can punt on the formatting for a follow up PR though to keep this one focused.

nplasterer (approved):
LGTM! Great work on this. I did a quick lint on the android files (removing unused imports and reformatting) and just pushed that in a commit.

Issue #66, opened 2023-07-04
Bug: sending wrong data to `createFromKeyBundle` makes app crash


nmalzieu:
### Describe the bug

on iOS, I sent the wrong data (JSONified UInt Array) to createFromKeyBundle and it crashed (it expects the key base64 encoded but it should not crash)

nplasterer:
Fixed by https://github.com/xmtp/xmtp-react-native/pull/70

PR #70, opened 2023-07-05
Fix: Refactors createFromKeyBundle swift function to guard string conv

Jonathansoufer:
This fixes the issue #66 when the app crashes while dealing with non String types on `createFromKeyBundle`.

**Taken Actions:**

- Added guard to the function to transfer program control out of scope.
- Added new error enum `invalidKeyBundle`.
- Returning empty string in case of problems.

**How to test it?**

- Pass any non string value to  `createFromKeyBundle` func.

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -143,8 +143,7 @@ class XMTPModule : Module() {
             clients[client.address] = client
             client.address
         } catch (e: Exception) {
-            Log.e("createFromKeyBundle", "Failed to create client: $e")
-            ""
+            throw XMTPException("Failed to create client: $e")
             }
         }
 
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -133,6 +133,7 @@ class XMTPModule : Module() {
         }
 
         AsyncFunction("createFromKeyBundle") { keyBundle: String, environment: String ->
+        try {
             logV("createFromKeyBundle")
             val options =
                 ClientOptions(api = apiEnvironments[environment] ?: apiEnvironments["dev"]!!)
@@ -141,6 +142,10 @@ class XMTPModule : Module() {
             val client = Client().buildFromBundle(bundle = bundle, options = options)
             clients[client.address] = client
             client.address
+        } catch (e: Exception) {
+            Log.e("createFromKeyBundle", "Failed to create client: $e")
+            ""
+            }
         }
 
         AsyncFunction("exportKeyBundle") { clientAddress: String ->
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -148,7 +148,7 @@ public class XMTPModule: Module {
                 return client.address
             } catch {
                 print("ERRO! Failed to create client: \(error)")
-                return ""
+                throw error
             }
         }
 
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -91,7 +91,8 @@ public class XMTPModule: Module {
     var subscriptions: [String: Task<Void, Never>] = [:]
 
     enum Error: Swift.Error {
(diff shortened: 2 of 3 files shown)


nplasterer:
Thanks so much for doing this!! I think it might be worth writing a test for this error maybe something like the below in this file https://github.com/xmtp/xmtp-react-native/blob/main/example/src/tests.ts . I ran this test below on main and saw the crash and also on your branch and saw it pass so can confirm this fixes it.

```js
import { randomBytes } from "crypto";

test("createFromKeyBundle throws error for non string value", async () => {
  try {
    const bytes = randomBytes(32)
    await XMTP.Client.createFromKeyBundle(JSON.stringify(bytes), "local");
  } catch (e) {
    return true;
  }
  return false;
});
```

Unfortunately our CI times out 75% of the time so we have to just run the unit tests locally in iOS and Android before merging. But hopefully I can get a fix for that in soon. But just giving you heads up that CI may be red even if its not.

ios/XMTPModule.swift
```diff
@@ -135,11 +136,21 @@ public class XMTPModule: Module {
...
+                }
+                
+                let options = XMTP.ClientOptions(api: apiEnvironments[environment] ?? apiEnvironments["dev"]!)
+                let client = try await Client.from(bundle: bundle, options: options)
+                self.clients[client.address] = client
+                return client.address
+            } catch {
+                throw error
+                print("ERRO! Failed to create client: \(error)")
+                return ""
```
nplasterer: I don't think you'll be able to get to these line of code if your throw on line 150 correct?

ios/XMTPModule.swift
```diff
@@ -135,11 +136,21 @@ public class XMTPModule: Module {
...
-            let bundle = try PrivateKeyBundle(serializedData: Data(base64Encoded: Data(keyBundle.utf8))!)
-            let options = XMTP.ClientOptions(api: apiEnvironments[environment] ?? apiEnvironments["dev"]!)
-            let client = try await Client.from(bundle: bundle, options: options)
-            self.clients[client.address] = client
-            return client.address
+            do {
+                guard let keyBundleData = Data(base64Encoded: keyBundle),
+                    let bundle = try? PrivateKeyBundle(serializedData: keyBundleData) else {
+                    throw Error.invalidKeyBundle
+                }
```
nplasterer, line 143: Seems like a change like this should probably also be done on the android side.

That being said from the test above I was able to confirm that android fails gracefully and did not crash so it's probably fine to not make the change in Android.

ios/XMTPModule.swift
Jonathansoufer: Indeed. Good catch. Removed and submitted again.

nplasterer: I actually think you wanted to keep the throw here instead of the return empty string. It isn't successfully creating a client so it's better to error than to not. The throw fixes the crashing issue by providing information about how to correctly send data to this function. Aka base64 encode it before sending.

ios/XMTPModule.swift
Jonathansoufer: Although Android is gracefully failing, I added a try-catch block to guarantee we keep exceptions handled.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -141,6 +142,10 @@ class XMTPModule : Module() {
             val client = Client().buildFromBundle(bundle = bundle, options = options)
             clients[client.address] = client
             client.address
+        } catch (e: Exception) {
+            Log.e("createFromKeyBundle", "Failed to create client: $e")
+            ""
```
nplasterer: Similar comment to what I put here: https://github.com/xmtp/xmtp-react-native/pull/70#discussion_r1253540807

I actually think you want to keep the throw. If a client isn't created there will be lots of other issues. It's better to fail with a throw than to return an empty string.
```suggestion
            throw XMTPException("Failed to create client: $e")
```

Jonathansoufer: Make sense, since it replicates what we did on iOS. Good one. Changed already.

nplasterer (approved):
This looks great thanks for addressing all my comments!

Screenshot 2023-07-05 at 3 29 32 PM

Issue #69, opened 2023-07-04
Bug: `listBatchMessages` does not handle Date parameters


nmalzieu:
### Describe the bug

it seems the `listBatchMessagesmethod` has an issue with dates :
it expects the before and after to be `Date` objects but when I try to send a date it says `ERROR in listBatchMessages [Error: The 6th argument cannot be cast to type Double?`
also it supposedly returns `DecodedMessage` objects where `sent` is of type `Date` but here I actually have `decodedMessage.sent` = 709725751.7279999 (what does this represent? It’s not a timestamp - this message has been sent recently)

daria-github:
Closed, resolved with https://github.com/xmtp/xmtp-react-native/pull/71

nplasterer:
Release in https://github.com/xmtp/xmtp-react-native/releases/tag/v1.6.1

PR #71, opened 2023-07-06
Fix date issue

daria-github:
This PR fixes an issue with passing custom dates to `listBatchMessages` — root of the issue was the `.getTime()` function itself was being passed to the Swift & Kotlin modules instead of its returned value. 

Also updated this for `listMessages` which uses similar logic, added test and added date rendering to the UI.  

Once this gets merged, should close out https://github.com/xmtp/xmtp-react-native/issues/69

### DIFF
--- a/example/src/ConversationView.tsx
+++ b/example/src/ConversationView.tsx
@@ -109,9 +109,17 @@ export default function ConversationView({ route }: Props): JSX.Element {
       >
         {messages.map((message) => {
           return (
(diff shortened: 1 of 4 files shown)


nplasterer (approved):
This looks good to me. Thanks for adding a test. Just one note about semantic releases. Would probably make sense to start your commit with "fix:" so that it gets released.

nakajima (approved):

PR #65, opened 2023-06-27
Docs: First pass of README preparing for beta

jhaaaa:
- Just putting this here for now - preparing a README for the beta release.
- Based on discussion with Naomi, starting with the README from xmtp.js and we can pare down and adjust as necessary.

PR #76, opened 2023-07-13
Readme for beta

nakajima:
Dup of https://github.com/xmtp/xmtp-react-native/pull/65.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,12 @@ To learn more, see [Use a local cache](https://xmtp.org/docs/tutorials/performan
 
 ## Create a client
 
-A client is created with `Client.create(wallet: Signer): Promise<Client>` that requires passing in a connected wallet that implements the [Signer](src/types/Signer.ts) interface. The client will request a wallet signature in two cases:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,17 +55,329 @@ npm i @xmtp/react-native-sdk
...
+### List messages in a conversation with pagination
+
+It may be helpful to retrieve and process the messages in a conversation page by page.
+
+```tsx
+const conversation = await xmtp.conversations.newConversation(
+  '0x3F11b27F323b62B159D2642964fa27C46C841897'
+)
+
+for await (const page of conversation.messages({ limit: 25 })) {
```
nplasterer: React question @daria-github can you pass things as an object? Even if it is setup to handle items. So I'd expect this to be actually `conversation.messages(limit: 25)` but wasn't sure if the object was hurting anything?

nplasterer (approved):
This looks good from my perspective.

README.md
daria-github: you can definitely pass an object in general — e.g. if this method was expecting an options object with a limit key this would be the correct syntax —  but looking at this function specifically it seems to just take in   individual arguments where we'd want to say` conversation.messages(25) `instead.

PR #77, opened 2023-07-18
List batch messages takes query parameter

### LINKED ISSUE #68: Feature request: specify timestamps in `listBatchMessages`  for each topic

In the Javascript SDK xmtp-js, there is the `batchQuery` method. This method can take multiple topics, and *for each topic* I can specify a start / end timestamp. It is very useful to do a performant synchronisation of our users' messages.

The method `listBatchMessages` in the RN SDK can take multiple topics but a single start / end date for all topics. I would love to have the same behaviour as the `batchQuery`.

_No response_

_No response_

### Additional details

_No response_

nplasterer:
Fixes https://github.com/xmtp/xmtp-react-native/issues/68

This takes a list of query objects so each topic can have it's own pagination associated with it.

Matches the query type specified in `xmtp-js` https://github.com/xmtp/xmtp-js/blob/main/src/ApiClient.ts#L56-L74

Note: This is a breaking change.

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -3,6 +3,7 @@ package expo.modules.xmtpreactnativesdk
 import android.util.Base64
 import android.util.Base64.NO_WRAP
 import android.util.Log
+import com.google.gson.GsonBuilder
 import com.google.protobuf.kotlin.toByteString
 import expo.modules.kotlin.modules.Module
 import expo.modules.kotlin.modules.ModuleDefinition
@@ -15,6 +16,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
+import org.json.JSONObject
 import org.xmtp.android.library.Client
 import org.xmtp.android.library.ClientOptions
 import org.xmtp.android.library.Conversation
@@ -25,6 +27,7 @@ import org.xmtp.android.library.messages.EnvelopeBuilder
 import org.xmtp.android.library.messages.InvitationV1ContextBuilder
 import org.xmtp.android.library.messages.PrivateKeyBuilder
 import org.xmtp.android.library.messages.Signature
+import org.xmtp.android.library.messages.Pagination
 import org.xmtp.android.library.push.XMTPPush
 import org.xmtp.proto.keystore.api.v1.Keystore.TopicMap.TopicData
 import org.xmtp.proto.message.contents.Content.EncodedContent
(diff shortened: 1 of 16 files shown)


src/lib/Query.ts
```diff
@@ -0,0 +1,6 @@
+export type Query = {
+  startTime?: Date | undefined;
+  endTime?: Date | undefined;
+  contentTopic: string;
+  pageSize?: number | undefined;
```
dmccartney, line 5: I think these should be
`topic`, `limit`, `before` and `after`

dmccartney: we might also want to simplify the encode/decode of dates by using something like:
`beforeMillis` and `afterMillis` and expecting millisecond-timestamps (which are easier to toss around).

nplasterer: I was trying to keep this as close in line with what the js sdk does so it's easier for integrators moving between the two. I know that the mobile sdks have a bit of a different structure but I think it's more likely that react native users will be more familiar with our js sdk and I'd like to keep it as close as possible to that.

dmccartney: Ah, gotcha. In that case
- `XmtpModule.kt` is using the wrong field names (it's plucking "before", "after" etc from the JSON)
- we'll need to fix how we serialize/deserialize the `Date` values (`JSON.serialize` converts them to ISO strings by default, not integer timestamps)
- `pageSize` should be an `integer` not just a `number` (or we should convert before serializing)

nplasterer, line 5: This type matches the JS type.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -196,13 +198,54 @@ class XMTPModule : Module() {
...
             val afterDate = if (after != null) Date(after) else null
 
-            client.conversations.listBatchMessages(topics, limit, beforeDate, afterDate).map {
+            conversation.messages(limit = limit, before = beforeDate, after = afterDate)
+                .map {
+                    EncodedMessageWrapper.encode(it)
+                }
+        }
+
+        AsyncFunction("loadBatchMessages") { clientAddress: String, topics: List<String> ->
```
nplasterer, line 217: `loadMessages` - lists all the messages via the envelopesList query inside of the SDKs

`loadBatchMessages` - lists all the messages for a batch of topics with given pagination parameters. 

These are two different methods in the sdks and should be here as well. Previously both messages, and batch messages called the batch method.

nakajima (approved):

src/index.ts
```diff
@@ -112,23 +112,21 @@ export async function listMessages(
...
-    before?.getTime(),
-    after?.getTime()
-  );
+  const topics = queries.map((item) => {
+    return JSON.stringify({
+      limit: item.pageSize || 0,
+      topic: item.contentTopic,
+      after: item.startTime?.getTime() || 0,
+      before: item.endTime?.getTime() || 0,
+    });
```
nplasterer, line 123: This converts the javascript type to the JSON string that the SDKS expect.

PR #79, opened 2023-07-25
Support App Version

### LINKED ISSUE #7: Support more ClientOptions
https://github.com/xmtp/xmtp-js/blob/fb4883cf81ce340b4caf9f1ccb60340808e303e9/src/Client.ts#L168

nplasterer:
This is a breaking change

This reworks the way we are doing client options currently by allowing an object to be passed instead. This will make it easier for us to allow more options and brings it more inline with how js does it.

closes: https://github.com/xmtp/xmtp-react-native/issues/7

### DIFF
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -70,20 +70,28 @@ extension Conversation {
 }
 
 public class XMTPModule: Module {
-    var apiEnvironments = [
-        "local": XMTP.ClientOptions.Api(
-            env: XMTP.XMTPEnvironment.local,
-            isSecure: false
-        ),
-        "dev": XMTP.ClientOptions.Api(
-            env: XMTP.XMTPEnvironment.dev,
-            isSecure: true
-        ),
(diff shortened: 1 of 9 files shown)


ios/XMTPModule.swift
```diff
@@ -70,20 +70,28 @@ extension Conversation {
...
+            case "local":
+                return XMTP.ClientOptions.Api(
+                        env: XMTP.XMTPEnvironment.local,
+                        isSecure: false,
+                        appVersion: appVersion
+                    )
+            case "production":
+                return XMTP.ClientOptions.Api(
+                        env: XMTP.XMTPEnvironment.production,
+                        isSecure: false,
```
dmccartney: I think this should be `isSecure: true` (for both `dev` and `prod`)

dmccartney (approved):
LG (after `isSecure` note addressed)

ios/XMTPModule.swift
nplasterer: Nice catch!

PR #81, opened 2023-08-03
feat: revamp example app, implement prelim content types

dmccartney:
This revamps the example app and adds support and demo of standard content types.
Instead of attempting to use the JS codecs, this takes the approach of relying on the native SDK codecs and then passing a JSON edition of the decoded content back/forth to the JS layer.

This means that adding a new content type looks something like:
1. design the content type
2. implement the codec in the native iOS and Android SDKs
3. add some glue code to this RN SDK to make use of it

This means there will always be a native codec for supported content types. This is important for native-only contexts like handling push notifications.

Here's my rough TODO list for tracking this WIP:
- [x] add and run `prettier`
- [x] revamp the example app UI
- [x] introduce RN abstraction over content types
  - [x] js
  - [x] android
  - [x] ios
- [ ] add RN bindings for standard content types
  - [x] Text
    - [x] android
    - [x] iOS
  - [x] Reaction
    - [x] android
    - [x] iOS
  - [x] Attachment
    - [x] android
    - [x] iOS
  - [x] Reply
    - [x] android
    - [x] iOS

These will come in a later PR:
  - [ ] Composite
    - [ ] android
    - [ ] iOS
  - [ ] RemoteAttachment
    - [ ] android
    - [ ] iOS


src/lib/Conversation.ts
```diff
@@ -48,13 +52,16 @@ export class Conversation {
   }
 
   // TODO: support conversation ID
-  async send(content: any): Promise<string> {
+  async send(content: string | MessageContent): Promise<string> {
```
dmccartney, line 54: gist of the `send(...)` signature change.

src/XMTP.types.ts
```diff
@@ -1,25 +1,54 @@
...
+// Each of these corresponds to a codec supported by the native libraries.
+// This is a one-of or union type: only one of these fields will be present.
+export type MessageContent = {
+  text?: string;
+  unknown?: UnknownContent;
+  reply?: ReplyContent;
+  reaction?: ReactionContent;
+  attachment?: AttachmentContent;
+  remoteAttachment?: RemoteAttachmentContent;
+};
```
dmccartney, line 46: gist of the new message content.
We pass one of these to `.send(...)` and we also receive one whenever we receive a `DecodedMessage`.

nplasterer:
I'm not sure if you need the group chat codecs just yet. I think @nakajima might be doing a revamp on how group chat works. Not sure if these will still be relevant.

peterferguson:
> I'm not sure if you need the group chat codecs just yet. I think @nakajima might be doing a revamp on how group chat works. Not sure if these will still be relevant.

Hey wondering if there is any timeline for group chat features in RN lib?

nakajima:
> Hey wondering if there is any timeline for group chat features in RN lib?

@peterferguson Not yet.

nplasterer (approved):
All these changes look great. Could be fine to do the rest of the codecs in a follow up PR. Thanks for all the tweaks to the example app as well. :shipit:

dmccartney:
Here are some rough screenshots of the overhauled app UIs:

Screenshot 2023-08-05 at 2 10 19 AM
Screenshot 2023-08-05 at 2 09 11 AM

PR #87, opened 2023-08-23
feat: include topic on the DecodedMessage

dmccartney:
This adds the conversation `topic` to the `DecodedMessage` so that it can be associated with the corresponding conversation.

This finishes the leftover bits from https://github.com/xmtp/xmtp-react-native/pull/86

This depends on
- https://github.com/xmtp/xmtp-ios/pull/148
- https://github.com/xmtp/xmtp-android/pull/110


nplasterer (approved):
Code looks good. Will just need to bump the versions when they land.

PR #101, opened 2023-09-07
removed deprecated classifier (Gradle 8) from Android build

sullyAL:
Gradle 8 has removed the depricated property `classifier ` and instead we need to use `archiveClassifier` otherwise the build will fail

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -37,7 +37,7 @@ buildscript {
 
 // Creating sources with comments
 task androidSourcesJar(type: Jar) {
-  classifier = 'sources'
+  archiveClassifier = 'sources'
   from android.sourceSets.main.java.srcDirs
 }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,10 +35,21 @@ buildscript {
   }
 }
 
+// Check the Gradle version
+def gradleVersion = GradleVersion.current()
+def requiredGradleVersion = GradleVersion.version('8.0')
+
 // Creating sources with comments
-task androidSourcesJar(type: Jar) {
-  archiveClassifier = 'sources'
-  from android.sourceSets.main.java.srcDirs
+if (gradleVersion < requiredGradleVersion) {
+    task androidSourcesJar(type: Jar) {
+      classifier  = 'sources'
+      from android.sourceSets.main.java.srcDirs
+    }
+} else {
+    task androidSourcesJar(type: Jar) {
+      archiveClassifier = 'sources'
+      from android.sourceSets.main.java.srcDirs
+    }
 }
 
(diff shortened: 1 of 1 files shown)


sullyAL:
Added backward compatibility as well

android/build.gradle
```diff
@@ -35,10 +35,21 @@ buildscript {
...
 
+// Check the Gradle version
+def gradleVersion = GradleVersion.current()
+def requiredGradleVersion = GradleVersion.version('8.0')
+
 // Creating sources with comments
-task androidSourcesJar(type: Jar) {
-  classifier = 'sources'
-  from android.sourceSets.main.java.srcDirs
+if (gradleVersion < requiredGradleVersion) {
```
nplasterer: How about doing something like this instead?

```suggestion
if (GradleVersion.current() < GradleVersion.version('8.0')) {
```

android/build.gradle
```diff
@@ -35,10 +35,21 @@ buildscript {
   }
 }
 
+// Check the Gradle version
+def gradleVersion = GradleVersion.current()
+def requiredGradleVersion = GradleVersion.version('8.0')
```
nplasterer: These aren't used anywhere else so probably don't need to be defined. 

```suggestion
```

nplasterer (commented):
Thanks for opening a PR just one small comment!

android/build.gradle
sullyAL: Hi there 

Yeah that's probably better, usually I define them if the inline conditions get to long but in this case is short and understandable enough 

I have commited the changes!

Thank you

nplasterer (approved):
Thanks so much for doing this. We use semantic releases so to get a new npm package you'll need to make a commit start with `fix:` in this case. You can read more here: https://github.com/semantic-release/semantic-release#commit-message-format

If merged in it's current form no npm package will be created. But that's okay if you are willing to wait a day or so for some other changes we're working on to land.

sullyAL:
Thank you for merging it! 

I'll wait until other changes are pushed as well

Many thanks

nplasterer:
@sullyAL there should be a bot that tells you this 🤔  but this contribution will earn you a POAP.

Congrats, your important contribution to this open-source project has earned you a GitPOAP!


[**GitPOAP: 2023 XMTP Contributor**](https://www.gitpoap.io/gp/1042):


GitPOAP: 2023 XMTP Contributor GitPOAP Badge


Head to [gitpoap.io](https://www.gitpoap.io) & connect your GitHub account to mint!

Learn more about GitPOAPs [here](https://medium.com/gitpoap/whats-a-gitpoap-72f1f5778c6/).

PR #113, opened 2023-09-13
fix: the reaction codec crash

### LINKED ISSUE #107: Native Reaction Codec crash on malformed reactions
The iOS app crashes with Fatal error: Unexpectedly found nil while unwrapping an Optional value from here : https://github.com/xmtp/xmtp-ios/blob/main/Sources/XMTP/Codecs/ReactionCodec.swift#L59-L63
I have some reactions that don’t have a schema (it was not mandatory at the very beginning) and maybe one that doesn’t have  reference 
 ReactionSchema(rawValue: content.parameters["schema"] ?? "")! , if there is no schema, is ReactionSchema(rawValue: "")!  but “” is not in the ReactionSchema enum so it’s nil and so it’s  (nil)! which does not make sense
It shouldn’t crash but just ignore the message if it’s malformed

this line
action: ReactionAction(rawValue: content.parameters["action"] ?? "")!,
if “action” is not part of content.parameters, is equal to
 ReactionAction(rawValue: "")! which is equal to
nil!
which is forbidden and makes the whole app crash

nplasterer:
Fixes https://github.com/xmtp/xmtp-react-native/issues/107

| Android | iOS |
|--------|--------|
| Screenshot 2023-09-14 at 9 58 47 AM | Screenshot 2023-09-14 at 9 29 04 AM |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -95,7 +95,7 @@ repositories {
 dependencies {
   implementation project(':expo-modules-core')
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${getKotlinVersion()}"
-  implementation "org.xmtp:android:0.6.1"
+  implementation "org.xmtp:android:0.6.2"
   implementation 'com.google.code.gson:gson:2.10.1'
   implementation 'com.facebook.react:react-native:0.71.3'
   implementation "com.daveanthonythomas.moshipack:moshipack:1.0.1"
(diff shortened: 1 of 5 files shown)


neekolas (approved):

neekolas:
Sorry, I never made a PR to update the `docker-compose.yml` for this repo. Would be something a lot like https://github.com/xmtp/xmtp-ios/pull/154

neekolas (approved):

PR #118, opened 2023-09-16
+ Typescript Cleanup

ryuzaki01:
this will cleanup most of the typescript issue in the example folder

### DIFF
--- a/example/src/XmtpContext.tsx
+++ b/example/src/XmtpContext.tsx
@@ -1,4 +1,4 @@
-import {createContext, FC, useContext, useMemo, useState} from "react";
+import {createContext, FC, ReactNode, useContext, useMemo, useState} from "react";
 import * as XMTP from "xmtp-react-native-sdk";
 
 const XmtpContext = createContext<{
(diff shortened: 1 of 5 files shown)


example/src/XmtpContext.tsx
```diff
@@ -9,7 +9,7 @@ const XmtpContext = createContext<{
   setClient: () => {},
 });
 export const useXmtp = () => useContext(XmtpContext);
-export function XmtpContextProvider({ children }) {
+export const XmtpContextProvider: FC<any> = ({ children }) => {
```
rygine: i think `FC<PropsWithChildren>` would work better here

ryuzaki01: ahhh yes, I missed that..

example/src/XmtpContext.tsx
```diff
@@ -9,7 +9,10 @@ const XmtpContext = createContext<{
   setClient: () => {},
 });
 export const useXmtp = () => useContext(XmtpContext);
-export function XmtpContextProvider({ children }) {
+type Props = {
+  children: ReactNode
+}
+export const XmtpContextProvider: FC<Props> = ({ children }) => {
```
rygine, line 15: FYI, `PropsWithChildren` is an official export from `react`

rygine (approved):

PR #119, opened 2023-09-16
+ Fix Docker image on Mac M1

ryuzaki01:
This is to fix error when running docker compose
```
no matching manifest for linux/arm64/v8 in the manifest list entries
```


neekolas (approved):
Thanks for fixing this. IIRC some recent Docker update changed the default behavior for emulation on M1 Macs to make this necessary.

PR #120, opened 2023-09-19
fix: concurrent threading issue

### LINKED ISSUE #115: Bug: crash when creating multiple clients on iOS
### Describe the bug

I'm working on multi account on Converse and need to instantiate multiple xmtp clients when opening the app
I experienced some crashes in the swift code at different places when I did `createFromKeyBundle`, always EXC_BAD_EXEC:

Here:
 
Capture d’écran 2023-09-14 à 18 23 44

And also here:

Capture d’écran 2023-09-14 à 18 38 00

when settings codecs[codec.id] for instance.

To reproduce I just did 
```
for (let index = 0; index < 200; index++) {
            Client.createRandom();
            console.log(index);
          }
```
and I have the crash every time

Looks like a thread safety issue maybe? I tried forcing those lines on main thread and it works but honestly I have no idea why:
Capture d’écran 2023-09-14 à 18 38 42
Capture d’écran 2023-09-14 à 18 38 52

N.B.: it seems that I don't have the error if I never launch multiple `createFromKeyBundle` or `createRandom` in // (just by adding an `await` before each `createRandom` call) it's not ideal and also the SDK should never make the app crash so it's worth exploring

### Expected behavior

_No response_

_No response_

nplasterer:
Fixes https://github.com/xmtp/xmtp-react-native/issues/115

### DIFF
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -276,11 +276,21 @@ public class XMTPModule: Module {
             }
 
             let conversations = try await client.conversations.list()
-
-            return try conversations.map { conversation in
-                await conversationsManager.updateConversation(key: conversation.cacheKey(clientAddress), conversation: conversation)
-
-                return try ConversationWrapper.encode(conversation, client: client)
+            
+            return try await withThrowingTaskGroup(of: String.self) { group in
+                for conversation in conversations {
+                    group.addTask {
+                        await self.conversationsManager.updateConversation(key: conversation.cacheKey(clientAddress), conversation: conversation)
+                        return try await ConversationWrapper.encode(conversation, client: client)
+                    }
+                }
+                
+                var results: [String] = []
+                for try await result in group {
+                    results.append(result)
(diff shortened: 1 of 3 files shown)


neekolas (approved):

ios/XMTPModule.swift
```diff
@@ -70,10 +70,52 @@ extension Conversation {
...
+        // A method to update the conversations
+        func updateClient(key: String, client: XMTP.Client?) {
+            clients[key] = client
+        }
+
+        // A method to retrieve a conversation
+        func getClient(key: String) -> XMTP.Client? {
+            return clients[key]
+        }
+    }
```
nakajima, line 90: Instead of wrapping these objects, could we just keep the dictionaries and use the `updateValue` function:

```swift
await clients.updateValue(client, forKey: "some address")
``` 

then we can just use the normal subscript getters.

nakajima (approved):

PR #121, opened 2023-09-19
Simplify Wallet buttons

nakajima:
Simplify this code, there's no need to iterate.

### DIFF
--- a/src/lib/Client.ts
+++ b/src/lib/Client.ts
@@ -140,7 +140,7 @@ export type NetworkOptions = {
   /**
    * Specify which XMTP environment to connect to. (default: `dev`)
    */
-  env: keyof typeof Env;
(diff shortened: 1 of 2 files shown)


src/lib/Client.ts
```diff
@@ -140,7 +140,7 @@ export type NetworkOptions = {
   /**
    * Specify which XMTP environment to connect to. (default: `dev`)
    */
-  env: keyof typeof Env;
```
nplasterer, line 143: Can you remove the Env enum so it's not confusing above this line

```
export enum Env {
  local = 'local',
  dev = 'dev',
  production = 'production'
}
```

nplasterer (approved):

src/lib/Client.ts
nakajima: Yep done.

PR #140, opened 2023-10-16
Move the SDK into Production

nplasterer:
This PR updates the readme and moves the sdk from Beta to Production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ To learn more about XMTP and get answers to frequently asked questions, see [XMT
 
 ## Reference docs
 
-Access the `xmtp-react-native` client SDK [reference documentation](https://xmtp.github.io/xmtp-react-native/).
+Access the `xmtp-react-native` client SDK [reference documentation](https://xmtp.github.io/xmtp-react-native/modules.html).
 
 ## Example app
 
--- a/README.md
+++ b/README.md
@@ -315,7 +315,11 @@ All send functions support `SendOptions` as an optional parameter. The `contentT
 
 To learn more about content types, see [Content types with XMTP](https://xmtp.org/docs/concepts/content-types).
 
-For example, see the [Codecs](https://github.com/xmtp/xmtp-react-native/blob/main/src/XMTP.types.ts) available in `xmtp-react-native`.
+The SDK preregisters the following codecs:
+
+- For [Android](https://github.com/xmtp/xmtp-react-native/blob/main/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt#L43-L53), using these [source codecs](https://github.com/xmtp/xmtp-android/tree/main/library/src/main/java/org/xmtp/android/library/codecs).
+
+- For [iOS](https://github.com/xmtp/xmtp-react-native/blob/main/ios/Wrappers/DecodedMessageWrapper.swift#L35-L48), using these [source codecs](https://github.com/xmtp/xmtp-ios/tree/main/Sources/XMTP/Codecs).
 
 ```tsx
   await conversation.send({
--- a/README.md
+++ b/README.md
@@ -4,14 +4,7 @@
 
 This repo provides a package you can use to build with XMTP in a React Native or Expo app.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,7 +92,7 @@ Currently, network nodes are configured to rate limit high-volume publishing fro
 > **Important**  
 > If you are building a production-grade app, be sure to use an architecture that includes a local cache backed by an XMTP SDK.  
 
-To learn more, see [Use a local cache](https://xmtp.org/docs/tutorials/performance#use-a-local-cache).
+To learn more, see [Use local-first architecture](https://xmtp.org/docs/build/local-first).
```
jhaaaa, line 95: Has local-first architecture been built into the SDK - as with React, for example? Just curious!

README.md
```diff
@@ -319,38 +315,19 @@ All send functions support `SendOptions` as an optional parameter. The `contentT
 
 To learn more about content types, see [Content types with XMTP](https://xmtp.org/docs/concepts/content-types).
 
-Support for other types of content can be added by registering additional `ContentCodecs` with the `Client`. Every codec is associated with a content type identifier, `ContentTypeId`, which is used to signal to the client which codec should be used to process the content that is being sent or received.
-
-<!--not sure about the Codecs link - WDYT?-->
-
-For example, see the [Codecs](https://github.com/xmtp/xmtp-react-native/blob/main/src/lib/CodecRegistry.ts) available in `xmtp-react-native`.
+For example, see the [Codecs](https://github.com/xmtp/xmtp-react-native/blob/main/src/XMTP.types.ts) available in `xmtp-react-native`.
```
jhaaaa: Updated this link - previous link is now a 404. Is this link correct? Is it correct to call this link "Codecs"?

README.md
```diff
@@ -369,71 +346,7 @@ The keys returned by `exportKeyBundle` should be treated with the utmost care as
...
-4. Uncomment `classpath('com.google.gms:google-services:4.3.15')` in the top level of the example app's `build.gradle` file.
-
-5. Sync the gradle project.
-
-6. Replace `YOUR_SERVER_ADDRESS` in the `PullController.ts` file. If you're using the example notification server, it should be something like `YOURIPADDRESS:8080` since the Android emulator takes over localhost.
-
-7. Change the example app's environment to `production` in both places in `AuthView.tsx`.
-
-8. Replace `YOUR_FIREBASE_SENDER_ID` in the `PullController.ts` with your sender ID from Firebase.
+Check out the `push-notifications-example` branch and follow instructions [in the README](https://github.com/xmtp/xmtp-react-native/blob/push-notifications-example/README.md).
```
jhaaaa, line 353: Add this link - correct?

jhaaaa:
Thank you so much for this @nplasterer! 🙇🏻‍♀️ 

Added just a few tech writer-type questions - and I think the README will be ready to go once we are ready for production release.

README.md
nplasterer: No. We haven't added infrastructure to allow the react native app to use the react hooks yet. The local storage work will likely come with V3 at this point.

README.md
nplasterer: Sort of it's the content types that we automatically support in react native. React Native is the only SDK that default registers all the codecs. You can see which ones are supported here https://github.com/xmtp/xmtp-react-native/blob/main/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt#L43-L53 and here https://github.com/xmtp/xmtp-react-native/blob/main/ios/Wrappers/DecodedMessageWrapper.swift#L35-L48 where we pre register the codecs. It might be worth linking out to the android and ios codecs inside of the other repos as that is actually where the codec code lives so like here: https://github.com/xmtp/xmtp-android/tree/main/library/src/main/java/org/xmtp/android/library/codecs

README.md
nplasterer: Yes. 👍

nplasterer:
The Android Test CI is flaky feel free to merge this without it passing 🙏

PR #145, opened 2023-11-01
Update block to deny and add implicit consent for sending messages

nplasterer:
This addresses the items outlined here: https://github.com/orgs/xmtp/discussions/49#discussioncomment-7429217

Changes the word blocked to denied and adds implicit consent for sending a message to a conversation with unknown consent.

### DIFF
--- a/example/src/HomeScreen.tsx
+++ b/example/src/HomeScreen.tsx
@@ -87,7 +87,7 @@ function ConversationItem({ conversation, client }: { conversation: Conversation
             ({messages?.length} messages)
           </Text>
           <Button
-            title="Denied"
+            title="Deny"
             onPress={denyContact}
             disabled={
               getConsentState == "denied"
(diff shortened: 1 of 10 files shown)


example/src/HomeScreen.tsx
```diff
@@ -87,10 +87,10 @@ function ConversationItem({ conversation, client }: { conversation: Conversation
             ({messages?.length} messages)
           </Text>
           <Button
-            title="Block"
-            onPress={blockContact}
+            title="Denied"
```
rygine: should this be `Deny`?

src/lib/Conversation.ts
```diff
@@ -114,7 +114,7 @@ export class Conversation {
     }
   }
 
-  async consentState(): Promise<"allowed" | "blocked" | "unknown"> {
+  async consentState(): Promise<"allowed" | "denied" | "unknown"> {
```
rygine, line 117: i see this string union type defined twice (here and in `index.ts`). think it's worth it to create 1 type and reference that instead?

rygine (approved):
LGTM. left some comments, but approving to unblock.

src/lib/Conversation.ts
nplasterer: Good call.

PR #157, opened 2023-11-30
Update README.md

WilliamYWY:
**Issues:**

- Found out that the xmtp dependencies name in package-lock.json has been changed to '@xmtp/react-native-sdk'

**Changes**

- Change import statement from '@xmtp/xmtp-react-native' to '@xmtp/react-native-sdk'


nplasterer (approved):
Thanks for the catch!

Issue #109, opened 2023-09-12
preCreateIdentityCallback and preEnableIdentityCallback


nplasterer:
Add `preCreateIdentityCallback` and `preEnableIdentityCallback` that we use during the onboarding from the JS SDK

Will likely need updates to iOS and Android.

nmalzieu:
@nplasterer Adding this here since to us it's linked to the onboarding!
Right now the RN SDK the `canMessage` method is only available on the client instance i.e. after having connected. But during onboarding we check if the current wallet is on XMTP to change the UX, so would be nice to have `canMessage` accessible directly on the `Client` class (with parameter for the network)

Kalas777:
👍🦺

PR #155, opened 2023-11-27
Custom content types

nakajima:
Introduces the ability to use custom content types on iOS (Kotlin support to follow).

This PR does not replace any of the existing content types that already ship with SDK. Instead it changes `DecodedMessage#content` to be a function rather than a property. The function will either return the same values it does today or the decoded content from a JS content codec. 

If we dig this overall approach then we can look into replacing the hard coded codecs in a couple places, but I don't know what that would look like for codecs that require a native component (like attachments).

Basic usage can be seen in the tests:

- [Defining a custom content type](https://github.com/xmtp/xmtp-react-native/pull/155/files#diff-ad2a5f900212586906998f46d89003c169d2425c5ad8f5b5ba82fa53fe254478R10-R39)
- [Using the custom content type](https://github.com/xmtp/xmtp-react-native/pull/155/files#diff-ad2a5f900212586906998f46d89003c169d2425c5ad8f5b5ba82fa53fe254478R676-R700)

PR #158, opened 2023-11-30
add staticCanMessage method

kele-leanes:
**Description:**
This PR makes available the `staticCanMessage` method which is used on the native packages. 
This change resolves the comment in this [discussion](https://github.com/xmtp/xmtp-react-native/issues/109)

| iOS| Android |
|:--:|:--:|
|  |  |


peterferguson:
Nice been looking forward to this! Currently not having this is eating up alot of our airstack calls 😆

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -214,6 +214,17 @@ class XMTPModule : Module() {
...
         }
 
+        AsyncFunction("staticCanMessage") { peerAddress: String, environment: String, appVersion: String? ->
+            try {
+                logV("staticCanMessage")
+                val options = ClientOptions(api = apiEnvironments(environment, appVersion))
+                Client.canMessage(peerAddress = peerAddress, options = options)
+                true
+            } catch (e: Exception) {
+                throw XMTPException("Failed to create client: $e")
```
nplasterer: This will just print the name of the class "Exception.class" I think you want the actual error message which would be like this.

```suggestion
                throw XMTPException("Failed to create client: ${e.message}")
```

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -214,6 +214,17 @@ class XMTPModule : Module() {
             client.canMessage(peerAddress)
         }
 
+        AsyncFunction("staticCanMessage") { peerAddress: String, environment: String, appVersion: String? ->
+            try {
+                logV("staticCanMessage")
+                val options = ClientOptions(api = apiEnvironments(environment, appVersion))
+                Client.canMessage(peerAddress = peerAddress, options = options)
+                true
```
nplasterer: This will always return true. I think you want it to return what is returned from canMessage.

You should just be able to remove this line.

```suggestion
```

example/src/tests.ts
```diff
@@ -620,3 +620,11 @@ test("canManagePreferences", async () => {
   return true
 });
 
+test("is address in the XMTP network", async () => {
+  const alix = await XMTP.Client.createRandom({ env: "local" });
+
+  const isAddressAvailable = await XMTP.publicCanMessage(alix.address, 'local');
+
+  return isAddressAvailable
```
nplasterer: Seems like you should also add a test that it returns false for addresses not on the network as well because you had a bug in the android code that would always return true that this didn't catch.

example/src/tests.ts
```diff
@@ -620,3 +620,11 @@ test("canManagePreferences", async () => {
   return true
 });
 
+test("is address in the XMTP network", async () => {
+  const alix = await XMTP.Client.createRandom({ env: "local" });
+
+  const isAddressAvailable = await XMTP.publicCanMessage(alix.address, 'local');
```
nplasterer: This name `publicCanMessage` doesn't exist anywhere in the code? Are you sure this passes?

nplasterer (commented):
It appears your missing the place on Client you want to put this static method. Probably here: https://github.com/xmtp/xmtp-react-native/blob/main/src/lib/Client.ts#L120

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -214,6 +214,17 @@ class XMTPModule : Module() {
             client.canMessage(peerAddress)
         }
 
+        AsyncFunction("staticCanMessage") { peerAddress: String, environment: String, appVersion: String? ->
+            try {
+                logV("staticCanMessage")
+                val options = ClientOptions(api = apiEnvironments(environment, appVersion))
+                val canMessage = Client.canMessage(peerAddress = peerAddress, options = options)
+                canMessage
```
nplasterer: There isn't any reason to store this you can just run it like this and that will return correctly.

```suggestion
                Client.canMessage(peerAddress = peerAddress, options = options)
```

kele-leanes:
here the evidence for the new test passed

Screenshot 2023-11-30 at 20 45 47

nplasterer:
This looks really great thanks for all the modifications! I'm just holding off on approving because I want this PR to land first https://github.com/xmtp/xmtp-react-native/pull/155 as it makes a lot of formatting changes and I don't want to have any further merge conflicts. 🙏

nplasterer (approved):
I pulled in the large codec changes, fixed the linting issues, and tweaked a few small things. LGTM :shipit:

PR #169, opened 2023-12-06
Move React Native to latest User Preferences

nplasterer:
This moves RN on to the latest user preference encryption scheme.

This also adds exposing the `consentList` to the integrator so they can more easily manage and populate databases.

| iOS | Android |
|--------|--------|
|Screenshot 2023-12-06 at 1 39 22 PM|Screenshot 2023-12-06 at 1 38 41 PM|

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -9,6 +9,8 @@ import com.google.gson.JsonParser
 import com.google.protobuf.kotlin.toByteString
 import expo.modules.kotlin.modules.Module
 import expo.modules.kotlin.modules.ModuleDefinition
+import expo.modules.xmtpreactnativesdk.wrappers.ConsentWrapper
+import expo.modules.xmtpreactnativesdk.wrappers.ConsentWrapper.Companion.consentStateToString
 import expo.modules.xmtpreactnativesdk.wrappers.ContentJson
 import expo.modules.xmtpreactnativesdk.wrappers.ConversationWrapper
 import expo.modules.xmtpreactnativesdk.wrappers.DecodedMessageWrapper
@@ -557,20 +559,13 @@ class XMTPModule : Module() {
 
         AsyncFunction("consentList") { clientAddress: String ->
             val client = clients[clientAddress] ?: throw XMTPException("No client")
-            client.contacts.consentList.entries.map { "${it.key}: ${consentStateToString(it.value)}" }
+            client.contacts.consentList.entries.map { ConsentWrapper.encode(it.value) }
         }
     }
 
     //
     // Helpers
     //
-    private fun consentStateToString(state: ConsentState): String {
-        return when (state) {
-            ConsentState.ALLOWED -> "allowed"
-            ConsentState.DENIED -> "denied"
-            ConsentState.UNKNOWN -> "unknown"
-        }
-    }
 
     private fun findConversation(
         clientAddress: String,
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -548,7 +548,8 @@ class XMTPModule : Module() {
 
         AsyncFunction("refreshConsentList") { clientAddress: String ->
             val client = clients[clientAddress] ?: throw XMTPException("No client")
-            client.contacts.refreshConsentList()
+            val consentList = client.contacts.refreshConsentList()
+            consentList.entries.map { ConsentWrapper.encode(it.value) }
         }
 
         AsyncFunction("conversationConsentState") { clientAddress: String, conversationTopic: String ->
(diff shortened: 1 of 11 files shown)


alexrisch (approved):

src/index.ts
```diff
@@ -353,6 +353,20 @@ export function refreshConsentList(clientAddress: string) {
   XMTPModule.refreshConsentList(clientAddress)
 }
 
+export async function consentList(
+  clientAddress: string
+): Promise<Map<string, 'allowed' | 'denied' | 'unknown'>> {
```
nplasterer: @alexrisch I'm sure theres a better typescript way to do these enum like strings?

example/src/tests.ts
```diff
@@ -672,6 +672,23 @@ test('canManagePreferences', async () => {
...
 
+  const boConsentList = await bo.contacts.consentList()
+  await delayToPropogate()
+
+  if (boConsentList.size !== 1) {
+    throw new Error(`consent list for bo should 1 not ${boConsentList.size}`)
+  }
+
+  const boConsentListState = boConsentList.get(
+    `address-${alixConversation.peerAddress.toLowerCase()}`
```
nplasterer: The consent list were meant to be flexible allowing for things other than just addresses to be stored. For example blocking a `conversationId` or entire group chats etc... Because of this there is a bit of knowledge that needs to be understood about how we store the keys in the map. A prefix of address will match to this address being allowed or denied. cc @fabriguespe as this may be a good thing to document.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -552,17 +552,26 @@ class XMTPModule : Module() {
...
                 ?: throw XMTPException("no conversation found for $conversationTopic")
-            when (conversation.consentState()) {
-                ConsentState.ALLOWED -> "allowed"
-                ConsentState.DENIED -> "denied"
-                ConsentState.UNKNOWN -> "unknown"
-            }
+            consentStateToString(conversation.consentState())
+        }
+
+        AsyncFunction("consentList") { clientAddress: String ->
```
alexrisch, line 561: Do we need a follow up PR for the iOS side as well?

alexrisch (approved):
Quick question but approved

ios/XMTPModule.swift
```diff
@@ -512,19 +512,31 @@ public class XMTPModule: Module {
...
 			}
-			switch await conversation.consentState() {
-			case .allowed: return "allowed"
-			case .denied: return "denied"
-			case .unknown: return "unknown"
-			}
+            return consentStateToString(state: await conversation.consentState())
 		}
+
+        AsyncFunction("consentList") { (clientAddress: String) -> [String] in
```
nplasterer, line 522: @alexrisch iOS side is here.

alexrisch: I may have missed this when scrolling 😆

src/index.ts
alexrisch: TLDR: String literals are totally fine and I prefer them personally, how you've done this is totally fine

You can also add them as either an enum or string literal union.
You already have a string literal union, just not exportable
You can also create an enum that can generate a literal, so users can use whichever they prefer.

Full example: https://www.typescriptlang.org/play?#code/FAUwdgrgtgBAwgezAZ3AFwKKVgb2DAmAQQBsSEB3EAExgF4YByAQzMpsYBp9CARcAJY16TaoI7dCMAKpgA1mEojGEeYophGwAL7BgAen0wASiACOEAQCcQyGABUAngAdbAYysDnaGABYAdACMMAAUgQCcMAByCABuIFAARiBWMABMAAyZAJTAaC4g8EioYGgAMgJoKawiAAYAJDiIKOhY0Nq1ANx6bsU+Vcg+DCG9LaUAXEVj5ZXVJNn0AHwweFKjyAgkIP7kAOYjxejZ3drdwABmqm5oAkgw6+gVgyHA2ZMAsszOADyDnmC7ThTEozKpWVjLVaEdY+KBfERgEAUGCfH5-AQAoHNEEVMEQkIAbR4UgIBMYAzQXCYrHIVGojAAupISaTybY0GkqdjWth-KRaTQmcSSYYilAoOgYAgID4SBjCslaTBmGBaIlmLR0QCYCAthLSjAKAIyPdmBBUMqwDqrFYEFZhVJRWSKQBmLmHUptKD+ABCGoA+l6hSzWRTfFTGOrNWh-rtGcyCAzjsKbGgIFYrXDnCduqKAOICeJ2M1oBC9KDOLZVS20NwACxAbjkxZNWt2diNaDr0p8Ngs1gxu2V1GolVuYBqAgrdrQyDy7JCLDYdMYydFAElzjBzSk7NQkOSYHK5IUAETg1UIKCnmBt5BAxIymB9yw2OxT5wzwfAnnQefPblPV5fl2GoZMKUXKMYFiZgrFXboIMAzBeT9ahA2wZMeCAA

alexrisch (approved):

src/index.ts
cameronvoell: Agree with @alexrisch , your implementation is fine. Marginal improvement IMO would be to export a type that you can reuse like in this diff => https://github.com/xmtp/xmtp-react-native/compare/np/bump-pod-for-new-pppp...np/bump-pod-for-new-pppp-union-type?expand=1

Exporting an enum like in @alexrisch 's link is the most flexible if we want to err on the side of developer choice, but I think the slightly more opinionated string literal union type is the default here.

example/src/tests.ts
nplasterer: We moved this closer inline with JS and return a `ConsentListEntry` instead now

PR #170, opened 2023-12-06
Example App - Live Update Deny

### LINKED ISSUE #160: Example app: Live-update deny/allow

On the conversations page, the "Deny" and "Allow" buttons don't auto-update to reflect any changes on the UI. 

This should auto-update without requiring user action.

_No response_

### Additional details

_No response_

alexrisch:
Resolves https://github.com/xmtp/xmtp-react-native/issues/160 
Added request after denying the conversation to set state on the conversation item
Made wallet string selectable for easier developing
Fixed Android Button styling on home screen

https://github.com/xmtp/xmtp-react-native/assets/23103150/0d38903f-9c51-4f49-a371-ff607b021a6f

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -1,6 +1,6 @@
 import { NavigationContainer } from '@react-navigation/native'
 import React from 'react'
-import { Button } from 'react-native'
+import { Button, Platform } from 'react-native'
 import { QueryClient, QueryClientProvider } from 'react-query'
 
 import ConversationCreateScreen from './src/ConversationCreateScreen'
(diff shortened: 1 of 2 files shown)


cameronvoell (approved):
The live update of denied status works great, nice fix!

Color update on Android is helpful as well.

I noticed that the selectable prop does not seem to be working for me on Android (tested on emulator and device, neither worked for me). It does make the text copy-able on iOS though, so still an improvement. Perhaps we can look at adding a copy button in some follow up PR later, since that prop seems to have some issues on Android.

Nice work!

nplasterer (approved):
LGTM! If you're looking for an easy way to get the addresses when you create a wallet they show up in the console to make it easier for copying.

PR #172, opened 2023-12-07
Updated example app to react-native 0.71.14

### LINKED ISSUE #171: Bug: Xcode 15 fails to build Example app for iOS because of boost C++ bug
### Describe the bug

When Xcode 15   (latest stable as of 12/7/23) is installed, default build instructions for running the Example app on iOS result in the following error:

```bash
› Compiling react-native Pods/RCT-Folly » json.cpp

❌  (ios/Pods/boost/boost/container_hash/hash.hpp:131:33)

  129 | #else
  130 |         template <typename T>
> 131 |         struct hash_base : std::unary_function<T, std::size_t> {};
      |                                 ^ no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
  132 | #endif
  133 | 
  134 |         struct enable_hash_value { typedef std::size_t type; };

```

That error is covered in react-native issue here: https://github.com/facebook/react-native/issues/37748

That error is [fixed](https://github.com/facebook/react-native/issues/37748#issuecomment-1587791079) in react-native version "0.71.11" (Example app is on "0.71.8" as of 12/7/23)

If for some reason we cannot update the react-native version immediately, there is a workaround that can temporarily be added to example/ios/Podfile:

```
installer.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', '_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION']
      end
    end
```

### Expected behavior

README.md instructions for running the Example app with default tool should not result in an error.

Xcode 15 installed

```bash
npm install
cd example
npx pod-install
npm run ios
```

cameronvoell:
Fixes #171 

Note the changes in this PR are all a result of updating react-native version in `example/package.json` to `0.71.14` and running npm install + running the ios example app.

Unit tests all passing

For reference, RN Upgrade helper doesn't seem to show anything else relevant for us. => https://react-native-community.github.io/upgrade-helper/?from=0.71.8&to=0.71.14

CleanShot 2023-12-07 at 09 05 08

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -12,14 +12,14 @@ PODS:
     - ExpoModulesCore
   - EXFont (11.1.1):
     - ExpoModulesCore
-  - EXImageLoader (4.3.0):
+  - EXImageLoader (4.4.0):
     - ExpoModulesCore
     - React-Core
-  - Expo (48.0.20):
+  - Expo (48.0.21):
     - ExpoModulesCore
-  - ExpoDocumentPicker (11.5.4):
+  - ExpoDocumentPicker (11.7.0):
     - ExpoModulesCore
-  - ExpoImagePicker (14.3.2):
+  - ExpoImagePicker (14.5.0):
     - ExpoModulesCore
   - ExpoKeepAwake (12.0.1):
     - ExpoModulesCore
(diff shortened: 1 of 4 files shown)


cameronvoell:
@daria-github @nplasterer Are there any additional ways we would want to test a RN version update like this other than in the example app / unit tests?  Wondering if we have any formal or informal procedures for reaching out to integration clients for example.

alexrisch (approved):

nplasterer:
@cameronvoell is the concern that this would break integrators who don't support the latest RN version? This probably goes hand in hand with @rygine suggestion to make a beta channel for RN like we have in JS. Especially as we make more potentially breaking changes like this. I also want to note that we do [semantic releases](https://github.com/semantic-release/semantic-release) here so if you don't include a commit with `feat:` or `fix:` it won't cut a new release.

cameronvoell:
> @cameronvoell is the concern that this would break integrators who don't support the latest RN version? This probably goes hand in hand with @rygine suggestion to make a beta channel for RN like we have in JS. Especially as we make more potentially breaking changes like this. 

@nplasterer Yes this is the type of concern I was referencing, though I think this particular RN upgrade looks low risk, I figured it was a good time to ask. I like the idea of having a beta channel for releases like we how do in xmtp-js. Created an [issue here](https://github.com/xmtp/xmtp-react-native/issues/174) for adding beta releases, will take a look at whats required.

> I also want to note that we do [semantic releases](https://github.com/semantic-release/semantic-release) here so if you don't include a commit with `feat:` or `fix:` it won't cut a new release.

Ah perfect, I'll review the semantic release setup and let team know if anything seems unclear. For this fix, cutting a new release doesn't seem necessary. Thanks!

nplasterer (approved):

PR #176, opened 2023-12-13
Added thirdweb ConnectWallet button to LaunchScreen

### LINKED ISSUE #159: Example app: Re-add logic to log in on a physical device

Previously we had the ability to log in on a non-randomly generated wallet in our SDK example app. This issue is to re-add that functionality.  

Here's the PR where this was removed and will need to be re-added: https://github.com/xmtp/xmtp-react-native/pull/81 

It's potentially just a matter of bringing back AuthView and meshing it with the LaunchScreen. 

Note: since this uses the`@thirdweb-dev/react-native` package, you'll need to --force when installing (feel free to explore other tools if they're better but this one seemed to work for our needs).

This is useful for debugging wallet-specific issues that are hard to reproduce on fresh wallets. 

### Additional details

_No response_

cameronvoell:
Closes https://github.com/xmtp/xmtp-react-native/issues/159

## Description

This PR adds the ability to connect with accounts from wallet apps via the Thirdweb SDK. The only wallets that were consistently working for me were Rainbow and Metamask so those are the only two options for now. 

#### To Do before removing PR draft state:
- [x] Fix disconnect + re-connect flows for ConnectWallet button
- [x] Add `example/.env` for reading thirdweb API key
- [x] Update `example/README.md`

## Testing Steps:

1. Follow new `example/README.md` steps for obtaining a Thirdweb client id (not mandatory, just clears network errors from SDK when it tries to obtain visual assets)
   - See https://github.com/xmtp/xmtp-react-native/blob/cameronvoell/example-wallet-auth/example/README.md#configure-thirdweb-client-api
 2. On a device with either Metamask or Rainbow Wallet installed with at least one account, install Example app and click connect button
 3. Select either Rainbow or Metamask
 4. Should be re-directed to the selected app, click connect when prompted (sometimes takes 5+ seconds)
 5. After connecting you should be re-directed back to the Example app. There should be new options to use the connected wallet on either a dev or local network. Select one.
 6. You should then be re-directed to Connected wallet app and prompted to sign either an enable account prompt, or first a create account and then enable account prompt. 
 7. After signing necessary prompts, you should enter the Conversations screen.
 8. If you back out of the conversations Screen, your key information should be in the SAVED WALLET, and you can use that to re-enter. 
 9. You can disconnect the wallet app by clicking on the Connect button while it is connected and then clicking the red exit arrow icon. 

### Log in flow with Metamask or Rainbow via Thirdweb SDK

| Connect Button | Thirdweb Modal | Connect Prompt | Loading indicator | 
| -------------- | ------------------- | -------------- | ----------------- | 
|  |  |  |  | 

| Connected Wallet Button | Sign Prompt |
| -------------- | ------------------- |
|  | 

~~Note on `npm install --legacy-peer-deps` and @thirdweb-dev/react-native `expo` dependency version
As mentioned in the linked issue, re-adding the dependency for `@thirdweb-dev/react-native` means that we need to always run `npm install --force` when installing dependencies in the example directory (README.md already includes these instructions). The reason is that `@thirdweb-dev/react-native` specifies a version for expo dependency that is less than our current version. I left a comment on the thirdweb sdk repo to see if updating this dependency has blockers for any reason => https://github.com/thirdweb-dev/js/pull/1778#issuecomment-1854501613~~
No longer relevant, as thirdweb is compatible with our version of expo now.


nplasterer (approved):
LGTM. 👍

PR #177, opened 2023-12-13
feat: add preCreateIdentityCallback & preEnableIdentityCallback

kele-leanes:
**Description:**
This pull request introduces the implementation of `preCreateIdentityCallback` and `preEnableIdentityCallback` functionalities, addressing the specific requirements outlined in [this ticket](https://github.com/xmtp/xmtp-react-native/issues/109). The implementation leverages event handling, as it is the most suitable approach for executing JavaScript code from the native side.

To achieve this, the code ensures proper execution of the specified callbacks, and subsequently, an unsubscription process is initiated. This unsubscription is crucial to prevent inadvertent re-execution of the callback every time the Client is created.

Thorough testing has been conducted with this [Android package](https://github.com/xmtp/xmtp-android/pull/144) version and this [iOS package](https://github.com/xmtp/xmtp-ios/pull/201)

| iOS | Android |
|:--:|:--:|
|  |  |


cameronvoell (commented):
I was able to see the callbacks running locally by testing against your linked android and ios PRs. Everything worked well. This PR looks good to me. 

I think we should wait for the Android and iOS PRs to merged before we merge this one so that we can update the native modules in this PR for xmtp-react-native to run properly.

As soon as those PRs are approved and merged, I think we can merge this one as well. Nice work @kele-leanes !

nplasterer:
I'm getting an error when I try to run the app after pulling in main and bumping the versions.

```
 ERROR  Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

It might be related to @alexrisch hook changes?

nplasterer:
I was able to revert the hooks PR and validate the example was working prior to that. I'm going to merge this to unblock integrators and we can fix the example app in a follow up as I confirmed the SDK works as intended only the example is broken.

nplasterer (approved):

alexrisch:
> I'm getting an error when I try to run the app after pulling in main and bumping the versions.
> 
> ```
>  ERROR  Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> ```
> 
> It might be related to @alexrisch hook changes?

Created a PR which should hopefully fix these weird resolutions and the hook PR: https://github.com/xmtp/xmtp-react-native/pull/185 

TLDR: Resolutions are a pain, and multiple react versions were being used

PR #178, opened 2023-12-13
Hook Setup

alexrisch:
Setup for hooks in module

Step 1 of many towards parity with Web on the JS side

Added xmtp context into module and out from example app Added useClient hook to help manage client

Follow ups, adding codecs, processors, namespaces, validators Adding client close method in disconnect callback


nplasterer (approved):
Having these hook inside of the sdk will be a huge improvement!

PR #179, opened 2023-12-14
Make allow and deny awaitable

nplasterer:
Right now methods client.contacts.allow(addresses) & client.contacts.deny(addresses) are not awaitable
To provide a good experience we need to know if we updated the consent state successfully in the protocol, would it be possible to make them awaitable and return only if the protocol state was successfully updated?


alexrisch (approved):

cameronvoell (approved):
LGTM!

PR #183, opened 2023-12-15
fix: register codecs on create of clients in android

nplasterer:
We weren't registering the codecs in Android when creating the client which could cause issues with rendering codecs. This registers the codecs on client create


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -145,6 +145,7 @@ class XMTPModule : Module() {
             signer = reactSigner
             val options = ClientOptions(api = apiEnvironments(environment, appVersion))
             clients[address] = Client().create(account = reactSigner, options = options)
+            ContentJson.Companion
```
nplasterer, line 148: I'm not sure it's necessary to call Companion here it could be implicitly called without it. But I wanted to explicit that this is triggering the `init` function that registers the codecs.

alexrisch (approved):

PR #184, opened 2023-12-16
Revert "Hook Setup"

nplasterer:
Reverts xmtp/xmtp-react-native#178


nplasterer:
Just used this to validate the sdk was working and the example app is the only thing broken.

PR #185, opened 2023-12-16
Fix Resolutions & Move to Yarn

alexrisch:
Problem:
NPM forces us to jump through a number of hoops
We seen an issue when dealing with React peer dependencies but they do not get handled appropriately

This is actually also seen in new expo modules created through the create-expo-module cli

Solution:
Stop forcing dependencies and working around dependency resolution and simply use yarn

Changes:
Removed package-lock.json
updated docs to point user towards yarn
Fix readme for npm pod-install to npx pod-install

Fixes Issues with Hooks addition from this https://github.com/xmtp/xmtp-react-native/pull/178 & identified on here https://github.com/xmtp/xmtp-react-native/pull/177


rygine:
looks like this is using Yarn v1. why not Yarn v4?

nplasterer (approved):
Works perfectly! Yay for yarn. 🎉

cameronvoell (approved):
Agree that defaulting to yarn in this repo will simplify our expo modules + example app setup. PR looks ready to merge to me!

alexrisch:
> looks like this is using Yarn v1. why not Yarn v4?

I still only have v1 setup and it still seems to be the stable version: https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable 
I'll do some reading, v1 should allow us with the easiest migration from npm

rygine:
> > looks like this is using Yarn v1. why not Yarn v4?
> 
> I still only have v1 setup and it still seems to be the stable version: https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable I'll do some reading, v1 should allow us with the easiest migration from npm

v1 has been in maintenance mode since 1/2020. latest version now is 4.0.2, which requires corepack to be enabled.

just need to run `corepack enable`, then `yarn set version stable`. be sure to update the [nodeLinker](https://yarnpkg.com/configuration/yarnrc#nodeLinker) setting to `node-modules` to disable PnP mode.

that should hopefully just work. there are some GH actions updates as well. happy to sync on this if you'd like.

cameronvoell:
I propose we merge this with yarn v1 because it's the yarn setup linked to [from Expo docs](https://docs.expo.dev/get-started/installation/#recommended-tools), and we know it resolves a build issue that is currently merged on main, and the CI is working as is in this PR, and then we do a follow up issue for migrating from yarn v1 to yarn v4 => https://yarnpkg.com/migration/guide

That way we can resolve the build issue that is merged to main right now, without having to make sure that our expo setup and CLI has any issues with yarn v4. thoughts? @rygine @alexrisch

rygine:
> I propose we merge this with yarn v1

sounds good!

Issue #180, opened 2023-12-14
Bug: unsubscribeFromMessages cannot unsubscribe from the stream.


jingledongding:
### Describe the bug

When I exit the chat, I am unable to unsubscribe from the message stream using unsubscribeFromMessages

### Expected behavior

 return () => {
      if (conversation) {
        console.log(
          conversation.client.address,
          conversation.topic,
        );
        unsubscribeFromMessages(
          conversation.client.address,
          conversation.topic,
        );
      }
    };

kele-leanes:
hi @jingledongding 
Do you mind showing me how you're doing the subscribing? Are you trying to unsubscribe when the user is logged out? What do you mean by exit the chat?

jingledongding:
`export function useXmtpStream(
  conversation?: Conversation<NativeMessageContent>,
) {

  useEffect(() => {
    const streamMessages = async () => {
      if (!conversation) {
        return;
      }
      conversation.streamMessages(async message => {
        console.log(message);
      });
    };
    streamMessages();
    return () => {
      if (conversation) {
        unsubscribeFromMessages(
          conversation.client.address,
          conversation.topic,
        );
      }
    };
  }, [conversation]);
}`
I'm wondering if it's my code that's wrong. But when I open the chat multiple times, the console.log executes the same number of times as I open the chat window

nplasterer:
Should be fixed in https://github.com/xmtp/xmtp-react-native/pull/187. Feel free to reopen if you are still seeing the issue on the latest version.

PR #187, opened 2023-12-18
fix: unsubscribe message event

kele-leanes:
**Description:**
This pull request addresses the concerns raised in issue #180 by implementing the removal of the event subscription. The primary focus is on enhancing the reliability of unsubscribing from the message stream.

It is highly recommended to utilize the method returned by the `streamMessages` function for unsubscribing, as opposed to relying solely on the static `unsubscribeFromMessages`. This is crucial because the latter may not comprehensively handle unsubscription and might inadvertently impact multiple listeners.

For instance, consider the updated usage example below:

```typescript
export function useXmtpStream(conversation?: Conversation<any>) {
  useEffect(() => {
    const streamMessages = () => {
      if (!conversation) {
        return;
      }

      return conversation.streamMessages(async (message) => {
        console.log(message);
      });
    };

    const unsubscribe = streamMessages();

    return () => {
      if (conversation) {
        unsubscribe?.(); // Invoke the unsubscribe method
      }
    };
  }, []);
}
```

By calling the `unsubscribe` method obtained from `streamMessages`, we ensure a more comprehensive unsubscription process, mitigating potential issues related to multiple listeners.

### DIFF
--- a/src/lib/Conversation.ts
+++ b/src/lib/Conversation.ts
@@ -229,7 +229,7 @@ export class Conversation<ContentTypes> {
   ): () => void {
     XMTP.subscribeToMessages(this.client.address, this.topic)
     const hasSeen = {}
-    XMTP.emitter.addListener(
+    const messageSubscription = XMTP.emitter.addListener(
       'message',
       async ({
         clientAddress,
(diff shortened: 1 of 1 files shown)


alexrisch (approved):

nplasterer (approved):
Seems like a good change!

PR #189, opened 2023-12-18
Remove Manual Install Step

alexrisch:
Problem:
There are a number of manual steps required when integrating with React Native SDK

Would like to make sure that this is indeed a requirement or if we can smooth out integration steps

Solution:
Move dependency to podspec
Remove ReadMe step

Opening as draft to see how CI handles this change

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ yarn i @xmtp/react-native-sdk
 In the `ios` directory, update your `Podfile` file as follows:
 
 - Set this value: `platform :ios, '16.0'`. This is required by XMTP.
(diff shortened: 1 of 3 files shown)


cameronvoell (approved):
Example ios build works for me with this update, nice to remove an install step 👍

nplasterer (approved):
🎉 🥳  Love all these improvements!

alexrisch:
cc @nakajima for visibility, I am guessing since the builds were successful and tests were too the change should be fine and confident to merge, but do you have any gotcha's?

nplasterer:
I think this is okay to merge. Signaling to @fabriguespe that this step should no longer be needed.

PR #191, opened 2023-12-20
Fix decoding messages

nplasterer:
When sending a text message with an empty string the app would error as the string was considered falsy since it was empty. We now handle all text messages as they are supported by default.

In this process I also caught that custom content types would break on decode because the codec did not exist. We now error gracefully for unsupported content types allowing for the handling of them in app.

Screenshot 2023-12-20 at 11 34 03 AM

### DIFF
--- a/src/lib/DecodedMessage.ts
+++ b/src/lib/DecodedMessage.ts
@@ -86,11 +86,18 @@ export class DecodedMessage<ContentTypes = any> {
       const codec = this.client.codecRegistry[
         this.contentTypeId
       ] as JSContentCodec<ContentTypes>
-
+      if (!codec) {
+        throw new Error(
+          `no content type found ${JSON.stringify(this.contentTypeId)}`
+        )
+      }
       return codec.decode(encoded)
     } else {
       for (const codec of Object.values(this.client.codecRegistry)) {
-        if ('contentKey' in codec && this.nativeContent[codec.contentKey]) {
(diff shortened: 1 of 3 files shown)


src/lib/DecodedMessage.ts
```diff
@@ -86,18 +86,27 @@ export class DecodedMessage<ContentTypes = any> {
       const codec = this.client.codecRegistry[
         this.contentTypeId
       ] as JSContentCodec<ContentTypes>
-
+      if (!codec) {
+        throw new Error(
+          `no content type found ${JSON.stringify(this.contentTypeId)}`
+        )
+      }
```
nplasterer, line 93: Fix for custom content types.

src/lib/DecodedMessage.ts
```diff
@@ -86,18 +86,27 @@ export class DecodedMessage<ContentTypes = any> {
...
       return codec.decode(encoded)
     } else {
       for (const codec of Object.values(this.client.codecRegistry)) {
-        if ('contentKey' in codec && this.nativeContent[codec.contentKey]) {
+        if (
+          ('contentKey' in codec && this.nativeContent[codec.contentKey]) ||
+          this.nativeContent.hasOwnProperty('text')
```
nplasterer, line 99: Text is always a supported content type.

nakajima (approved):

PR #192, opened 2023-12-21
fix: callbacks in auth flow no longer unsubscribed too early

cameronvoell:
Callbacks for identity creation and enabling were added recently in the following PR:
- https://github.com/xmtp/xmtp-react-native/pull/177

Though that PR's automated tests verified the flow worked with randomly generated identity flow, we did not have a External Wallet SDK signing to test against to verify that the callbacks returned as expected.

In progress work on connecting external wallets revealed that during the regular auth flow, we were unsubscribing from the. new callbacks too early. 

The fix in this PR can be tested via the following test branch which has the fix in this PR as well as external wallet connect button: [callbacks-test-branch-with-wallet-connect](https://github.com/xmtp/xmtp-react-native/tree/callbacks-test-branch-with-wallet-connect)

In addition to moving the subscription removal to a later time, we also added error handling so that if a user rejects a signing request, that callback subscriptions are still removed as expected.

### DIFF
--- a/src/lib/Client.ts
+++ b/src/lib/Client.ts
@@ -62,22 +62,32 @@ export class Client<ContentTypes> {
           'sign',
           async (message: { id: string; message: string }) => {
             const request: { id: string; message: string } = message
-            const signatureString = await signer.signMessage(request.message)
-            const eSig = utils.splitSignature(signatureString)
-            const r = hexToBytes(eSig.r)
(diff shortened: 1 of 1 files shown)


cameronvoell:
@kele-leanes tagging you in case you see any issues with the PR here since you worked on this previously. 🙏

nplasterer (approved):
LGTM 👍

PR #198, opened 2024-01-02
Add Support for Custom Content Types when preparing messages

alexrisch:
Added logic in Conversation to handle content type in prepare method Updated bridged module to handle encrypted messages

Requires iOS Sdk Change: https://github.com/xmtp/xmtp-ios/pull/212
Requires Android Sdk Change: https://github.com/xmtp/xmtp-android/pull/150


nplasterer (approved):
This implementation looks correct to me. Once iOS and Android land this LGTM. 👍

PR #203, opened 2024-01-05
fix: set semaphore as nil after complete de function

kele-leanes:
**Description:**
This PR fixes the pre-event callbacks on iOS setting as nill only when the execution is finished. The issue happens when a user tries to log in with a wallet that has never been connected with XMTP. Triggering the `create` callback first and then waiting until `enable` ends the execution to launch the Enable signature.

**Evidence:**
_In this test a 10 sec timer was sent as a callback_
<video src='https://github.com/xmtp/xmtp-react-native/assets/57466680/013a8a73-edc2-4d6d-a144-5d359afb4871' width='600' />


nplasterer (approved):

nplasterer:
I think you'll need to push an empty commit with "fix: " in it so this triggers a release when merged.

PR #207, opened 2024-01-10
feat: Added support for Java 17

alexrisch:
Fix issue:

> 'compileDebugJavaWithJavac' task (current target is 17) and 'compileDebugKotlin' task (current target is 11) jvm target compatibility should be set to the same Java version.

Seen when working with React Native 0.73 project

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -66,15 +66,18 @@ afterEvaluate {
 }
 
 android {
-  compileSdkVersion safeExtGet("compileSdkVersion", 31)
+  compileSdkVersion safeExtGet("compileSdkVersion", 34)
 
-  compileOptions {
-    sourceCompatibility JavaVersion.VERSION_11
-    targetCompatibility JavaVersion.VERSION_11
-  }
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() < 8) {
+    compileOptions {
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -68,13 +68,16 @@ afterEvaluate {
 android {
   compileSdkVersion safeExtGet("compileSdkVersion", 31)
```
nplasterer: With the bump to 17 for the java version we can probably bump the compile version to the latest as well.
```suggestion
  compileSdkVersion safeExtGet("compileSdkVersion", 34)
```

android/build.gradle
```diff
@@ -68,13 +68,16 @@ afterEvaluate {
...
+      sourceCompatibility JavaVersion.VERSION_17
+      targetCompatibility JavaVersion.VERSION_17
+    }
 
-  kotlinOptions {
-    jvmTarget = JavaVersion.VERSION_11.majorVersion
+    kotlinOptions {
+      jvmTarget = JavaVersion.VERSION_17.majorVersion
+    }
   }
```
nplasterer, line 81: We want compileOptions and kotlinOptions to always be set. I actually think you should get rid of the if statement here. Seems like that could introduce an issue where they some how are not set.

alexrisch: I think it's expected from a React Native perspective: https://github.com/react-native-community/discussions-and-proposals/issues/671

nplasterer (approved):

PR #208, opened 2024-01-10
Added test for custom type encoding

cameronvoell:
We are seeing an issue where `encodedContent` in Custom types is unexpectedly base64 encoded which does not match the behavior on xmtp-js. 

In Addition, the test in this PR shows that encodedContent.content is received as undefined, only on Android. 

The goal in this PR is to fix the two issues caught by the new test `register and use custom content type transaction reference`


cameronvoell:
Similar test added in https://github.com/xmtp/xmtp-react-native/pull/209, and issue resolved

PR #209, opened 2024-01-11
Fix Custom Content Type Decryption

nplasterer:
- Adds content to the json blob return from android
- Converts base64 string (which is required since we cant base arrays across the native bridge) back to byte array

To pass data from native iOS and Android to RN we parse the information into json strings. To make Android match iOS we have to hand roll a lot of this. Because our test wasn't thorough I missed the content field in the json blob initially. We also missed the step to convert the base64 back into bytes after coming across the wire.

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
@@ -51,8 +51,6 @@ class ContentJson(
             Client.register(RemoteAttachmentCodec())
             Client.register(ReplyCodec())
             Client.register(ReadReceiptCodec())
-            // TODO:
-            //Client.register(CompositeCodec())
         }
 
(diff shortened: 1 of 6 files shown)


cameronvoell (approved):
Worked for me, nice fix! 

The buffer package is pretty [lightweight](https://bundlephobia.com/package/buffer@6.0.3) and [well-supported](https://github.com/feross/buffer) so there shouldn't be any issues adding it to the rn sdk. 

As a minor size improvement, we may be able to replace the 'buffer' dependency with [base64-js ](https://www.npmjs.com/package/base64-js)which is a dependency in `buffer` that is used for base64 decoding([link](https://github.com/beatgammit/base64-js)) and has its own method for converting from base64 string to byte array.

An alternative would be [react-native-quick-base64](https://github.com/craftzdog/react-native-quick-base64) if we ever want to aim for performance improvements, say if decoding large files in custom content is every a thing, but that package is less popular. 

I think a good path would be to merge with the buffer dependency to get this working and evaluate less popular alternatives as potential improvements 👍

PR #210, opened 2024-01-13
Fixes extra sign listeners

cameronvoell:
When testing against the Client.create() method we were seeing extra Error messages that hinted at native module listeners not being removed when clients were created subsequent time. 

After the code update, we no longer see the extra log error messages.


nplasterer (approved):
LGTM!

PR #215, opened 2024-01-22
chore: remove ethers dependency

### LINKED ISSUE #186: Feature request: Remove Ethers as a dependency

Hey glad to see this package getting so much love over the past few weeks!

Since there are more people working on the package now I was wondering whether there would be capacity to remove `ethers` as a dependency. This is currently to the last library we have that depends explicitly on it & we would really like to remove it from our bundle.

It is really only used [here](https://github.com/xmtp/xmtp-react-native/blob/c51dbbabbb99c044b32e1919758a4eedb460efec/src/lib/Client.ts#L65C9-L73C1) & what it is being used for could easily be broken out into a util in this code base. Or it could even be pushed to the user (probably not ideal).

The benefit here would be allowing the user to decide which type of `Signer` (`viem`, `ethers`, ...) they pass to the `Client` as long as it has a `signMessage` method.

_No response_

_No response_

### Additional details

_No response_

kele-leanes:
Closes https://github.com/xmtp/xmtp-react-native/issues/186

**Description:**
This PR aims to remove the `ethers` dependency from the sdk to allow the user to use a custom Signer as long as the `getAddress` and `signMessage` methods are available.
Only the sub-module [bytes](https://www.npmjs.com/package/@ethersproject/bytes) from ethers was added to still have access to the utils methods.

| iOS |
|:--:|
| <video src='https://github.com/xmtp/xmtp-react-native/assets/57466680/bb9ed759-1e56-4eb8-9b24-9bad91936088' width='400' /> |


src/lib/Signer.ts
```diff
@@ -0,0 +1,41 @@
+import type { WalletClient } from 'viem'
```
cameronvoell, line 1: I'm not sure if we want to add 'viem' as a dependency in the xmtp-react-native SDK. Just like how developers who use 'viem' wont want 'ethers' as a dependency, developers who use 'ethers' dont need to have 'viem' as a dependency either. 

I recognize that the `convertWalletClientToSigner` is a nice convenience feature, but it seems to go against the spirit of the [issue](https://github.com/xmtp/xmtp-react-native/issues/186) we are tackling.

Maybe we could remove the `viem` import in the SDK part of the code, and instead add a test to Example project that shows how you can create a Signer interface using a viem WalletClient?

kele-leanes: I see your point but consider that `viem` is a dev dependency. it's only importing a type.

I don't want to add more noise to the LaunchScreen but I'll try to do it as simply as possible. Thanks for your comment 😃

cameronvoell: Ah good point, you're right, if we're just importing types and adding as a dev dependency, it wont add to bundle size.  👍 In that case no update needed. Might just update the syntax on the other import so it also shows explicitly we're just importing the type. 👌

src/lib/Client.ts
```diff
@@ -1,5 +1,6 @@
-import { Signer, utils } from 'ethers'
+import { splitSignature } from '@ethersproject/bytes'
 import { Subscription } from 'expo-modules-core'
+import { WalletClient } from 'viem'
```
cameronvoell:
```suggestion
import type { WalletClient } from 'viem'
```

cameronvoell (approved):
Nice update!

PR #217, opened 2024-01-23
fix: Revert Java Version Upgrade

alexrisch:
Revert java version upgrade until React Native 0.73 is more widly used

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -66,17 +66,17 @@ afterEvaluate {
 }
 
 android {
-  compileSdkVersion safeExtGet("compileSdkVersion", 34)
+  compileSdkVersion safeExtGet("compileSdkVersion", 33)
 
   def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
   if (agpVersion.tokenize('.')[0].toInteger() < 8) {
     compileOptions {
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -66,17 +66,17 @@ afterEvaluate {
 }
 
 android {
-  compileSdkVersion safeExtGet("compileSdkVersion", 34)
+  compileSdkVersion safeExtGet("compileSdkVersion", 31)
```
nplasterer: We can bump this to 33 at the very least
```suggestion
  compileSdkVersion safeExtGet("compileSdkVersion", 33)
```

android/build.gradle
```diff
@@ -66,17 +66,17 @@ afterEvaluate {
...
 
   def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
   if (agpVersion.tokenize('.')[0].toInteger() < 8) {
```
nplasterer, line 72: This call is strange to me cause you have to be on AGP 8 for version 17. 🤔  But now that we are dropping it back to 11 not an issue. Maybe what needed to happen was an else statement that set to 17 if above AGP 8 otherwise 11.

nplasterer (approved):
We can revisit how to rework this in a way that works for older clients.

Issue #213, opened 2024-01-22
Tracking: RN Group Chat - Alpha


nplasterer:
### RN Android

- [ ] Reproduce utility classes/methods
  - [ ] Client
    - [x] update existing `create` method on client to also create a `FfiXmtpClient`
    - [ ] Existing client should now also have a `FfiLogger`
    - [x] Have existing SigningKey extend `FfiInboxOwner` so it can be used to create both V2 and V3 clients
    - [x] If `FfiXmtpClient` exists get address from `account_address()`
    - [ ] If `FfiXmtpClient` exists store a reference to `installation_id()`
    - [ ] Conversations on client should also take a `libXMTPConversations`
    - [x] If `FfiXmtpClient` exists client create should call `register_identity`
    - [ ] Ability to `list_my_installations()`
    - [ ] Ability to `revoke_installation()`
  - [ ] Conversations
    - [x] Should be able to create a group `newGroup()` -- `create_group()`
    - [x] Should be able to list just groups `listGroups` -- `libXMTPConversations.list()`
    - [x] Should be able to list an ordered list of conversations & groups `list(includeGroups: Boolean = false)`
    - [x] Should be able to stream groups `libXMTPConversations.stream()`
    - [x] Should be able to stream both conversations & groups `stream(includeGroups:Boolean = false)`
    - [x] Should be able to `sync` groups so that the libxmtp database gets updated
    - [x] Should be able to stream all messages from both groups and conversations (`streamAllMessages(includeGroups: Boolean = false)`
  - [ ] Conversation/Group methods
    - [x] Should have a unique identifier of the conversation `id()` equivalent to existing topic
    - [ ] Should be able to `send()` messages of any content type
    - [x] Should be able to resync the libxmtp database `sync()`
    - [x] Should be able to list messages `find_messages()` as Decoded and Decrypted Messages
    - [x] Should be able to list members `list_members()`
    - [x] Should be able to add members `add_members()`
    - [x] Should be able to remove members `remove_members()`
    - [x] Should be able to stream messages `stream()`
    - [ ] Should be able to leave a group `leave()`
    - [x] Can check if a message can be sent `can_message()`
- [ ] Publish repo README covering Dev Preview features and with Dev Preview badge

### RN iOS

- [ ] Reproduce utility classes/methods
  - [ ] Client
    - [x] update existing `create` method on client to also create a `FfiXmtpClient`
    - [ ] Existing client should now also have a `FfiLogger`
[...]

PR #231, opened 2024-01-31
Create LibXMTP Client - android

cameronvoell:
See https://github.com/xmtp/xmtp-react-native/issues/213

This change includes basic group chat functionality on Android:
- Create a group using a list of addresses
- list groups for a client
- refresh group list
- list members of a group
- message a group (default text message codec only)
- refresh messages for a group


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -76,16 +79,25 @@ class ReactNativeSigner(var module: XMTPModule, override var address: String) :
...
             continuations[request.id] = continuation
         }
     }
 
-    override suspend fun sign(message: String): Signature =
-        sign(message.toByteArray())
+    override fun sign(message: String): ByteArray {
+        return runBlocking {
+            signLegacy(message).toByteArray()
+        }
```
nplasterer: 🤔 I wonder if this will work for sure worth trying but I almost think that would just give me the bytearray of the signature object

When I think we want something more like this 
https://github.com/xmtp/xmtp-android/pull/156/files#diff-2be5e4bd7739dcb1debb7023943ee5544ebef6355405115cf89aa0e0a3ee09edR97-R102

Maybe this would do to get the bytes out of the signature 🤔 

```suggestion
        return runBlocking {
            signLegacy(message).ecdsaCompact.bytes.toByteArray()
        }
```

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -180,7 +200,7 @@ class XMTPModule : Module() {
         }
 
         // Generate a random wallet and set the client to that
-        AsyncFunction("createRandom") { environment: String, appVersion: String?, hasCreateIdentityCallback: Boolean?, hasEnableIdentityCallback: Boolean? ->
+        AsyncFunction("createRandom") { environment: String, appVersion: String?, hasCreateIdentityCallback: Boolean?, hasEnableIdentityCallback: Boolean?, enableAlphaMls: Boolean? ->
```
nplasterer, line 202: Maybe you could put a check at the top that errors if `enableAlphaMls` is true and the env isn't local?

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -224,6 +252,13 @@ class XMTPModule : Module() {
             }
         }
 
+        Function("getLibXMTPClientAccountAddress") { clientAddress: String -> String
+            logV("getLibXMTPClientAccountAddress")
+            val client = clients[clientAddress] ?: throw XMTPException("No client")
+            val libXMTPClient = client.libXMTPClient ?: throw XMTPException("No libxmtp client")
+            return@Function libXMTPClient.accountAddress()
+        }
```
nplasterer: You don't need this is handled on the backend now. Should be the same as the address that gets returned from the client.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -192,22 +212,30 @@ class XMTPModule : Module() {
...
             ContentJson.Companion
             clients[randomClient.address] = randomClient
             randomClient.address
         }
 
-        AsyncFunction("createFromKeyBundle") { keyBundle: String, environment: String, appVersion: String? ->
+        AsyncFunction("createFromKeyBundle") { keyBundle: String, environment: String, appVersion: String?, enableAlphaMls: Boolean? ->
             try {
                 logV("createFromKeyBundle")
-                val options = ClientOptions(api = apiEnvironments(environment, appVersion))
```
nplasterer, line 210: Same here for the check on local env if mls enabled

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -163,10 +179,14 @@ class XMTPModule : Module() {
                 preCreateIdentityCallback.takeIf { hasCreateIdentityCallback == true }
             val preEnableIdentityCallback: PreEventCallback? =
                 preEnableIdentityCallback.takeIf { hasEnableIdentityCallback == true }
+            val context = if (enableAlphaMls == true) context else null
```
nplasterer, line 181: And also here since we only support local.

nplasterer (commented):
This is looking really good just a few small comments

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
cameronvoell: Added 01e69b9bdda21ec07b8cc23ef7a08c2ca7459ca0

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
cameronvoell: Added 01e69b9bdda21ec07b8cc23ef7a08c2ca7459ca0

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
cameronvoell: Added 01e69b9bdda21ec07b8cc23ef7a08c2ca7459ca0

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -531,6 +611,47 @@ class XMTPModule : Module() {
...
+            if (client.libXMTPClient == null) {
+                throw XMTPException("Create client with enableAlphaMLS true in order to create a group")
+            }
+            val group = client.conversations.newGroup(peerAddresses)
+            logV("id after creating group: " + Base64.encodeToString(group.id, NO_WRAP))
+            val encodedGroup = GroupWrapper.encode(client, group)
+            return@AsyncFunction encodedGroup
+        }
+
+        AsyncFunction("listMembers") { clientAddress: String, groupId: String ->
```
nplasterer: I might change this to `listMemberAddresses` in case we do create some sort of member object down the road

cameronvoell: fixed here 19e03d84fa660046ebb6dae777fe3811f2b22868

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
cameronvoell: fixed here 573b2afd0e07b64aefed47897c260e8a008d47d4

alexrisch (approved):

PR #232, opened 2024-02-01
Identity signer

nplasterer:
This exposes a function to sign messages from identities. Will help unlock frames.
Also adds a way to fetch the public key bundle.


neekolas:
I realized there's actually one more thing I'll need to make this work. It looks like RN doesn't have a way to get the `PublicKeyBundle` for the current user. We'll also need that to assemble the Frame post payload

nplasterer:
I think they do but it's named differently https://github.com/xmtp/xmtp-react-native/blob/main/src/lib/Client.ts#L305-L315 is this what you need? @neekolas

neekolas:
That gets the PrivateKeyBundle, which I guess does included the public key bundle. Would be nice to not have to touch the private keys, but that would be fine as a short term fix. 

Maybe a few weeks from now we end up re-implementing all of this in Rust and we don't have to worry about it.

nplasterer:
Ahh okay I can definitely get the public keys easy then.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -224,6 +225,25 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("sign") { clientAddress: String, message: List<Int> ->
```
neekolas: I think we assume the message is already hashed in `PrivateKey.sign`, so we should probably call the input `digest` instead of `message`.

neekolas: Perfectly fine for me to send over digests. That's how I'm doing it in JS anyways

src/index.ts
```diff
@@ -76,6 +76,29 @@ export async function createFromKeyBundle(
...
+    Array.from(digest),
+    keyType,
+    preKeyIndex
+  )
+  return new Uint8Array(signatureArray)
+}
+
+export async function exportPublicKeyBundle(
+  clientAddress: string
+): Promise<Uint8Array> {
```
nplasterer, line 97: Not sure whats more helpful returning a bytearray or a string? I notice we return a string for the keybundle below.

neekolas: Byte array is good, since we are going to need to reinflate it to a full proto object later. And that process expects a byte array.

alexrisch:
Simulator Screenshot - iPhone 15 Pro - 2024-02-14 at 17 40 14

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -224,6 +227,36 @@ class XMTPModule : Module() {
...
+            val signature = runBlocking {
+                PrivateKeyBuilder(key).sign(digestBytes)
+            }
+            signature.toByteArray().map { it.toInt() and 0xFF }
+        }
+
+        AsyncFunction("exportPublicKeyBundle") { clientAddress: String ->
+            logV("exportPublicKeyBundle")
+            val client = clients[clientAddress] ?: throw XMTPException("No client")
+            client.privateKeyBundle.toV2().toPublicKeyBundle().toByteArray().map { it.toInt() and 0xFF }
```
neekolas: What's this `and 0xFF` business, just for my own understanding

alexrisch:
Screenshot 2024-02-15 at 3 49 42 PM

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
alexrisch: I think it's a way of dropping some bit from java bit handling, Naomi would know more than me

src/lib/Client.ts
```diff
@@ -408,6 +421,11 @@ export type NetworkOptions = {
   appVersion?: string
 }
 
+export type KeyType = {
```
alexrisch, line 424: Should move this to types folder

example/src/tests.ts
```diff
@@ -925,3 +940,40 @@ test('correctly handles lowercase addresses', async () => {
...
+    buttonIndex: 1,
+    conversationTopic: 'foo',
+    participantAccountAddresses: ['amal', 'bola'],
+  })
+  const postUrl = metadata.extractedTags['fc:frame:post_url']
+  const response = await framesClient.proxy.post(postUrl, signedPayload)
+  if (!response) {
+    throw new Error('response should exist')
+  }
+  console.log('response', response)
```
neekolas: You mind removing the console.log

example/package.json
```diff
@@ -14,6 +14,7 @@
     "@react-navigation/native-stack": "^6.9.12",
     "@thirdweb-dev/react-native": "^0.6.2",
     "@thirdweb-dev/react-native-compat": "^0.6.2",
+    "@xmtp/frames-client": "^0.3.0",
```
neekolas: 0.3.2 is now available

neekolas (approved):

src/lib/Client.ts
alexrisch: Actually there's a few other types defined there, will hold off on pattern change for now

cameronvoell (approved):
tests look good, nice work!

Issue #230, opened 2024-01-31
Bug: Case Insensitive Address Checks


alexrisch:
### Describe the bug

Creating a conversation with lowercase address that is on the network causes an error to occur

### Expected behavior

Should create a conversation regardless of casing, Related web issue: https://github.com/xmtp/xmtp-js/issues/472

nplasterer:
I recently learned that checksum addresses is an extra level of security https://eips.ethereum.org/EIPS/eip-55 which is why addresses created are neither all lowercase or all uppercase. We can't really control what they create a client with. The address form that is. So definitely some improvement on the sdk side to not error on casing.

cameronvoell:
See comment here - https://github.com/xmtp/xmtp-react-native/pull/233#issuecomment-1922439276 Might want to check if this needs to be fixed for consent, deny function etc.

alexrisch:
Included in version 1.26.0

PR #233, opened 2024-02-01
feat: Added support for handling lowercase addresses

alexrisch:
Will now get checksum addresses from provided addresses Added some helper libraries to avoid adding additional dependency on ethers

Simulator Screenshot - iPhone 15 Pro - 2024-02-01 at 14 10 37

### DIFF
--- a/src/utils/address.ts
+++ b/src/utils/address.ts
@@ -1,0 +1,50 @@
+import { keccak_256 } from '@noble/hashes/sha3'
+import { TextEncoder } from 'text-encoding'
+
+const addressRegex = /^0x[a-fA-F0-9]{40}$/
+const encoder = new TextEncoder()
+
+export function stringToBytes(value: string): Uint8Array {
+  const bytes = encoder.encode(value)
+  return bytes
+}
+
+export function keccak256(value: Uint8Array): Uint8Array {
+  const bytes = keccak_256(value)
+  return bytes
+}
+
+export function isAddress(address: string): boolean {
+  return addressRegex.test(address)
+}
+
+export function checksumAddress(address_: string, chainId?: number): string {
+  const hexAddress = chainId
+    ? `${chainId}${address_.toLowerCase()}`
+    : address_.substring(2).toLowerCase()
+  const hash = keccak256(stringToBytes(hexAddress))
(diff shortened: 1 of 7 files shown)


src/utils/address.ts
```diff
@@ -0,0 +1,50 @@
+import { keccak_256 } from '@noble/hashes/sha3'
```
alexrisch, line 1: Port from viem getAddress

cameronvoell:
this change looks good, sounds like it is addressing a specific issue with  adding new conversations.

I wonder if we should add the new getAddress in a few more locations. For example if you pass in a lowercase address to client.contacts.deny in the test linked below, I think we will fail to block the address. Perhaps we should merge this change, but keep your [issue](https://github.com/xmtp/xmtp-react-native/issues/230) open until we feel we have a comprehensive list of places where checksums need to added. WDYT?

https://github.com/xmtp/xmtp-react-native/blob/c68bcc551ab7ec1eb3de8ae6ec936fb6eaff4172/example/src/tests.ts#L676

cameronvoell (approved):
Approved with the recommendation we keep https://github.com/xmtp/xmtp-react-native/issues/230 open for now

alexrisch:
> this change looks good, sounds like it is addressing a specific issue with adding new conversations.
> 
> I wonder if we should add the new getAddress in a few more locations. For example if you pass in a lowercase address to client.contacts.deny in the test linked below, I think we will fail to block the address. Perhaps we should merge this change, but keep your [issue](https://github.com/xmtp/xmtp-react-native/issues/230) open until we feel we have a comprehensive list of places where checksums need to added. WDYT?
> 
> https://github.com/xmtp/xmtp-react-native/blob/c68bcc551ab7ec1eb3de8ae6ec936fb6eaff4172/example/src/tests.ts#L676

Good catch, let me figure out LOE and determine

alexrisch:
> Approved with the recommendation we keep #230 open for now

Updated handling for contacts methods

cameronvoell (approved):
Changes look good, think this PR should fully close https://github.com/xmtp/xmtp-react-native/issues/230 now, nice!

PR #244, opened 2024-02-07
Groups streaming (android)

cameronvoell:
Adds Group function `streamGroupMessages()` which should function the same as the existing `conversation.streamMessages()`

See new test:  https://github.com/xmtp/xmtp-react-native/pull/244/files#diff-ad2a5f900212586906998f46d89003c169d2425c5ad8f5b5ba82fa53fe254478R865-R924


nplasterer (approved):
Looks great!

PR #245, opened 2024-02-07
Group Messages Decoding

alexrisch:
Decode Group Messages


nplasterer (approved):

cameronvoell:
thanks for the fix on this @alexrisch ! 🙇‍♂️  I just updated the test to use the proper decoded fields.

PR #249, opened 2024-02-09
Fix client create for groups

nplasterer:
Also missing the flag here

Also pulls main into beta.


alexrisch (approved):

cameronvoell (approved):
thanks for adding the missing flags 🙏

PR #250, opened 2024-02-09
Add/remove groups

alexrisch:
Added functionality for add remove groups

Screenshot 2024-02-09 at 9 43 11 AM

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -636,6 +636,28 @@ class XMTPModule : Module() {
             runBlocking { group?.sync() }
         }
 
+        AsyncFunction("addGroupMembers") { clientAddress: String, id: String, peerAddresses: List<String> ->
+            logV("addGroupMembers")
+            val client = clients[clientAddress] ?: throw XMTPException("No client")
+            if (client.libXMTPClient == null) {
+                throw XMTPException("Create client with enableAlphaMLS true in order to create a group")
+            }
+            val group = findGroup(clientAddress, id)
+
+            runBlocking { group?.addMembers(peerAddresses) }
+        }
+
+        AsyncFunction("removeGroupMembers") { clientAddress: String, id: String, peerAddresses: List<String> ->
+            logV("removeGroupMembers")
+            val client = clients[clientAddress] ?: throw XMTPException("No client")
+            if (client.libXMTPClient == null) {
+                throw XMTPException("Create client with enableAlphaMLS true in order to create a group")
+            }
+            val group = findGroup(clientAddress, id)
+
+            runBlocking { group?.removeMembers(peerAddresses) }
+        }
+
         Function("subscribeToConversations") { clientAddress: String ->
             logV("subscribeToConversations")
             subscribeToConversations(clientAddress = clientAddress)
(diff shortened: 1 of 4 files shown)


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -636,6 +636,28 @@ class XMTPModule : Module() {
...
 
+        AsyncFunction("addGroupMembers") { clientAddress: String, id: String, peerAddresses: List<String> ->
+            logV("addGroupMembers")
+            val client = clients[clientAddress] ?: throw XMTPException("No client")
+            if (client.libXMTPClient == null) {
+                throw XMTPException("Create client with enableAlphaMLS true in order to create a group")
+            }
+            val group = findGroup(clientAddress, id)
+
+            runBlocking { group?.addMembers(peerAddresses) }
```
richardhuaaa, line 647: In the long-term it might be good to return an error if the group is not found, here and in the other methods, but doesn't need to block this PR

richardhuaaa (approved):
Good job on the fast turnaround!

example/src/tests.ts
```diff
@@ -154,58 +159,279 @@ test('can message in a group', async () => {
...
+  // await bobClient.conversations.syncGroups()
+
+  // bobGroups = await bobClient.conversations.listGroups()
+  // await bobGroups[0].sync()
+  // const bobGroupMessages = await bobGroups[0].messages()
+  // if (bobGroups.length !== 0) {
+  //   throw new Error(
+  //     'num groups for bob should be 0, but it is ' + bobGroups.length
+  //   )
+  // }
```
cameronvoell, line 429: I noticed that when this is uncommented that the test fails because the size of `bobClient.conversations.listGroups()` is still 1 after he is removed.

@richardhuaaa Is this expected that groups you have been removed from still are returned from listGroups()? If so, what is the proper way for someone to detect that they have been removed from a group?

richardhuaaa: Hey @cameronvoell, yes with the current code the group and message history will remain in your database, but you should stop receiving updates. @neekolas is working on a way to notify the platform/app layer that a user has been removed from a group, but it is not ready yet.

> Is this expected that groups you have been removed from still are returned from listGroups()?

Alex was also confused about this, lmk if this seems unintuitive from the perspective of API design or if there's a better way to expose this. I don't think we've necessarily optimized around developer experience so far

alexrisch: For now it's expected, a user will see past messages but not future

alexrisch: @richardhuaaa I think from a UX perspective it's a bit confusing
From a developer perspective it makes sense if the purpose is to allow the developer freedom to clear the messages in some way.
Meaning if I as an integrating developer want to "remove" the user from the chat and clear messages I can, or if I want them to just stop receiving updates I have the options to do either

cameronvoell: > Hey cameronvoell, yes with the current code the group and message history will remain in your database, but you should stop receiving updates. neekolas is working on a way to notify the platform/app layer that a user has been removed from a group, but it is not ready yet.

> Alex was also confused about this, lmk if this seems unintuitive from the perspective of API design or if there's a better way to expose this. I don't think we've necessarily optimized around developer experience so far 

@richardhuaaa Sounds like it's just unintuitive because that one piece is not finished yet. Think the only real missing piece is a recommended way for someone to be notified that theyve been blocked, once we have that we should be able to handle the rest on the SDK side. Thanks for confirming!

neekolas: Agreed. I don't think it's our job to decide what to do _after_ the user has been removed. Some apps may want to continue showing the conversation and history (just with the ability to interact removed), some won't. We just need to surface that information to developers. I'll work on adding that today.

PR #257, opened 2024-02-13
fix: updated some generics for groups

alexrisch:
Update generics for groups and client creation


cameronvoell (approved):
thanks for updating!

PR #262, opened 2024-02-14
Remove base64 for group id

nplasterer:
Lets just convert the group id to hex like we do everywhere else in the code.


alexrisch (approved):

cameronvoell (approved):
LGTM I added one more spot for base64 to hex 👍

PR #273, opened 2024-02-20
Added tests for capturing createdAt values and list function behavior

cameronvoell:
This PR adds tests for checking the order returned from `list()`, `listGroups()`, and `listAll()`, `stream()`, `streamGroups()`, and `streamAll()`, and compares `createdAt` field of items returned from list functions to `createdAt` field of items returned by `createGroup/Conversation` functions.

My recommendation is to merge these new tests with the comments they have about where behavior is currently inconsistent between android and iOS. We can then update the tests as the below issues are fixed. 

| Android | iOS |
| ------------- | ------------- |
|    |   |

Issues Documented in the tests:

listGroups issues: See https://github.com/xmtp/xmtp-react-native/issues/276
1. `conversations.listGroups()` returns in reverse chronological on iOS, and chronological order on Android
2. `createdAt` field returned by` listGroups()` does not match the `createdAt` of group returned by `createGroup` function on Android when the group was created by someone other than the client who called `listGroups()`

list / listAll issues: See https://github.com/xmtp/xmtp-react-native/issues/277
1. `conversations.listAll()` sets values to `createdAt` for groups that do not match the values of `createdAt` returned from `createGroup` function.
2. `list()` and `listAll()`  methods for conversations returns in Chronological order for iOS and reverse Chronological order for Android

stream/streamAll issue: See https://github.com/xmtp/xmtp-react-native/issues/278
- Both stream() and streamAll() the `createAt` value of a group returned does not match the `createdAt` value of the group returned by the createGroup() method on Android only.

### DIFF
--- a/example/src/tests/createdAtTests.ts
+++ b/example/src/tests/createdAtTests.ts
@@ -1,6 +1,4 @@
-import { Platform } from 'expo-modules-core'
-
-import { Test, assert, delayToPropogate } from './tests'
+import { Test, assert, delayToPropogate, isIos } from './tests'
 import {
   Client,
   Conversation,
@@ -9,10 +7,6 @@ import {
 } from '../../../src/index'
 
 export const createdAtTests: Test[] = []
-
-function isIos() {
-  return Platform.OS === 'ios'
-}
 
 function test(name: string, perform: () => Promise<boolean>) {
   createdAtTests.push({ name, run: perform })
@@ -166,7 +160,6 @@ test('group createdAt matches streamGroups', async () => {
   const cancelStream = await aliceClient.conversations.streamGroups(
     async (group: Group<any>) => {
       allGroups.push(group)
-      console.log('group', allGroups.length)
     }
   )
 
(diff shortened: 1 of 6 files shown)


example/src/tests/createdAtTests.ts
```diff
@@ -0,0 +1,548 @@
...
+  const bobGroup = await bobClient.conversations.newGroup([aliceClient.address])
+
+  // Fetch groups using listGroups method
+  await aliceClient.conversations.syncGroups()
+  const aliceGroups = await aliceClient.conversations.listGroups()
+
+  // BUG - List returns in Reverse Chronological order on iOS
+  // and Chronological order on Android
+  const first = isIos() ? 1 : 0
+  const second = isIos() ? 0 : 1
```
alexrisch, line 46: Do we have a tracking issue for this?

alexrisch (approved):

example/src/tests/createdAtTests.ts
cameronvoell: https://github.com/xmtp/xmtp-react-native/issues/276, and there's two more issues linked in the description 👍

PR #274, opened 2024-02-20
Default Sync for group methods

### LINKED ISSUE #263: Default groups list methods to sync, optional parameter for skipping sync

_No response_

_No response_

_No response_

### Additional details

_No response_

cameronvoell:
Closes #263 

All tests are still passing even though all `group.sync()` and `conversation.syncGroups()` calls have been removed from tests.

### DIFF
--- a/src/lib/Group.ts
+++ b/src/lib/Group.ts
@@ -38,7 +38,10 @@ export class Group<
     return this.client.address
   }
 
-  async memberAddresses(): Promise<string[]> {
+  async memberAddresses(skipSync = false): Promise<string[]> {
+    if (!skipSync) {
+      await this.sync()
+    }
     return XMTP.listMemberAddresses(this.client, this.id)
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/Group.ts
```diff
@@ -76,7 +79,10 @@ export class Group<
     }
   }
 
-  async messages(): Promise<DecodedMessage<ContentTypes>[]> {
+  async messages(skipSync = false): Promise<DecodedMessage<ContentTypes>[]> {
```
alexrisch, line 82: Thoughts on using an object instead?
Could get confusing especially when adding pagination and sorting later on

alexrisch (approved):
Approving so I'm not blocking, but added a question

src/lib/Group.ts
cameronvoell: Yea I agree (if im understanding correctly), once we have pagination and sorting, passing in an object would be more straightforward. My inclination is to wait until one of those is ready before adding here, so I'll merge for now with the idea we'll update when those are ready in upcoming PR 🙏

PR #275, opened 2024-02-20
Group Admin & Message Pagination

nplasterer:
This pulls in the latest of the SDK versions which has
- [x] Group Admin Base functionality
- [x] Fixes to Can Message
- [x] Group message pagination

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -184,7 +184,7 @@ class XMTPModule : Module() {
             val preEnableIdentityCallback: PreEventCallback? =
                 preEnableIdentityCallback.takeIf { hasEnableIdentityCallback == true }
             val context = if (enableAlphaMls == true) context else null
-            
+
             val options = ClientOptions(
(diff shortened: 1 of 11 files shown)


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -466,14 +463,20 @@ class XMTPModule : Module() {
                 .map { DecodedMessageWrapper.encode(it) }
         }
 
-        AsyncFunction("groupMessages") { clientAddress: String, id: String ->
+        AsyncFunction("groupMessages") { clientAddress: String, id: String, limit: Int?, before: Long?, after: Long?, direction: String? ->
             logV("groupMessages")
             val client = clients[clientAddress] ?: throw XMTPException("No client")
-            if (client.libXMTPClient == null) {
-                throw XMTPException("Create client with enableAlphaMLS true in order to synch groups")
-            }
```
nplasterer, line 474: I made `libXMTPClient` private in the android sdk like iOS so that integrators can't call the rust library directly without going through the sdk protections. This means we can no longer do this check in RN. But that should be oaky because it will throw from Android if they don't have it setup.

cameronvoell (approved):

PR #280, opened 2024-02-21
Stream All Group Messages

nplasterer:
Adds the ability to stream all group messages and to stream all messages including groups.

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -673,9 +673,9 @@ test('can stream group messages', async () => {
 })
 
 test('can stream all messages', async () => {
-  const bo = await Client.createRandom({ env: 'local' })
+  const bo = await Client.createRandom({ env: 'local', enableAlphaMls: true })
   await delayToPropogate()
-  const alix = await Client.createRandom({ env: 'local' })
+  const alix = await Client.createRandom({ env: 'local', enableAlphaMls: true })
   await delayToPropogate()
 
   // Record message stream across all conversations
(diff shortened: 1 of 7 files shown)


example/src/tests/groupTests.ts
```diff
@@ -800,3 +803,60 @@ test('can paginate group messages', async () => {
...
+
+  await alix.conversations.streamAllGroupMessages(async (message) => {
+    allMessages.push(message)
+  })
+
+  for (let i = 0; i < 5; i++) {
+    await boGroup.send({ text: `Message ${i}` })
+    await delayToPropogate()
+  }
+  if (allMessages.length <= 10) {
```
cameronvoell, line 857: This test was failing for me here. ~~Think it's supposed to be greater than, to verify that cancellation worked?~~
Hmm. Actually it's only failing for me on Android. Digging in a bit more now, might just need a delay for a slow emulator.

example/src/tests/groupTests.ts
```diff
@@ -800,3 +803,60 @@ test('can paginate group messages', async () => {
...
+    await caroGroup.send({ text: `Message ${i}` })
+    await delayToPropogate()
+  }
+
+  if (allMessages.length !== 10) {
+    throw Error('Unexpected all messages count second' + allMessages.length)
+  }
+
+  alix.conversations.cancelStreamAllGroupMessages()
+
```
cameronvoell: Figured out that this test only passes for me on Android if I have an additional delay between the`cancelStreamAllGroupMessages()` and starting the stream again. Otherwise `allMessages.length` would equal 10 for me in the last check. 

```suggestion
  await delayToPropogate()
```

cameronvoell (approved):

PR #294, opened 2024-02-29
fix: Additional checks on stream callbacks

alexrisch:
Added checks for ids on JS callbacks
Added tests for streams with mutliple clients

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -93,7 +93,9 @@ test('can make a MLS V3 client with encryption key and database path', async () 
   await client.conversations.newGroup([anotherClient.address])
   assert(
     (await client.conversations.listGroups()).length === 1,
-    `should have a group size of 1 but was ${(await client.conversations.listGroups()).length}`
+    `should have a group size of 1 but was ${
+      (await client.conversations.listGroups()).length
+    }`
   )
 
   const bundle = await client.exportKeyBundle()
@@ -112,7 +114,9 @@ test('can make a MLS V3 client with encryption key and database path', async () 
 
   assert(
     (await clientFromBundle.conversations.listGroups()).length === 1,
-    `should have a group size of 1 but was ${(await clientFromBundle.conversations.listGroups()).length}`
+    `should have a group size of 1 but was ${
+      (await clientFromBundle.conversations.listGroups()).length
+    }`
   )
   return true
 })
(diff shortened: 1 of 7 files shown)


example/src/tests/groupTests.ts
```diff
@@ -932,3 +936,300 @@ test('can stream all group messages', async () => {
...
   return true
 })
+
+test('can streamAll from multiple clients', async () => {
+  const bo = await Client.createRandom({ env: 'local', enableAlphaMls: true })
+  await delayToPropogate()
+  const alix = await Client.createRandom({ env: 'local', enableAlphaMls: true })
+  await delayToPropogate()
+  const caro = await Client.createRandom({ env: 'local', enableAlphaMls: true })
+  await delayToPropogate()
```
cameronvoell, line 991: heads up - I added a helper method for client creation to make the tests less lines of code. It's been working fine in my tests without `delayToPropogate` in between, but feel free to modify `test-utils/createClients.ts` if you see error on client creation ( so far I have not).

```suggestion
const [alix, bo, caro] = await createClients(3)
```

cameronvoell: We can always clean up all the tests at once with the helper method in a separate PR if you prefer, but figured I'd let you know since it makes writing/reading tests a little easier.

src/lib/Group.ts
```diff
@@ -133,13 +133,19 @@ export class Group<
       async ({
         clientAddress,
         message,
+        groupId,
       }: {
         clientAddress: string
         message: DecodedMessage<ContentTypes>
+        groupId: string
       }) => {
+        // Long term these checks should be able to be done on the native layer as well, but additional checks in JS for safety
```
cameronvoell, line 142: I was thinking about this as well. My first instinct was to filter on native side, but I think instead we can use some shared EventEmitter listeners for each event type on the JS side that is shared between clients. 

Currently, any time we call Group.streamGroupMessages() or conversation.streamMessages(), we are registering an additional Expo Native Module [EventEmitter](https://docs.expo.dev/modules/module-api/#sending-events) Listener that will receive all events of type 'message' or 'groupMessage'.

Ideally we would only ever have one active EventEmitter listener per Event type, and some way to manage adding and removing callback functions targeting specific conversation topics, group ids and specific clients. Let me know if that idea sounds right to you. I think this is an improvement that would remove unnecessary extra messages being passed to listeners on the JS side that need to be weeded out, but is not essential for correct behavior.

cameronvoell (approved):
One comment with a suggestion on using a helper function for tests, and another on a potential improvement I'd be interested to hear your opinion on. 

I think this addresses some of the immediate issues with messages showing up unexpectedly in streams. Nice fix + tests 👍

src/lib/Group.ts
alexrisch: Personally I think it would be better to have handling for multiple listeners/callbacks
When callbacks get registered all of the callbacks work

I think one issue is right now we can't fire "dynamic" events, they have to be predefined first

But normal RN modules I think should be fine 🤔

example/src/tests/groupTests.ts
alexrisch: Yep, saw your PR was waiting for it to merge I'll update to that

src/lib/Group.ts
cameronvoell: Good point, we can get around the pre-registered event emitters, if we move way from expo native modules. Agree that would work as well 👌

alexrisch: Hold me to it, expo removal PR by next week

example/src/tests/groupTests.ts
alexrisch: Thank you for the example by the way, made it very easy to update without needing to track down additional code

PR #295, opened 2024-03-02
Rust iOS Streaming Updates

nplasterer:
This should fix some of the streaming issues we've been seeing.

Should we also bring this into the main branch?
This means iOS and Android will diverge a bit again. But the android streaming stuff is going to be a little more complicated and lengthy fix so I think this would still be good to land on iOS in the meantime.


neekolas (approved):

nplasterer:
Okay looks like there is a legit test failure with returning iOS messages. These are the only things that changed in iOS https://github.com/xmtp/xmtp-ios/pull/270. Seems likely it's related to the latest rust update.

PR #296, opened 2024-03-06
feat: get HMAC keys method

kele-leanes:
TBD


alexrisch:
iOS is working and validated
Android there is an issue on the native layer, but the bridging seems correct
We can merge this to start integrator validation now and allow some additional time for Android to get fixed

alexrisch (approved):

Issue #306, opened 2024-03-11
Bug: streamMessages and streamAllMessages stop streaming message after short period of time


WilliamYWY:
### Describe the bug

streamMessages and streamAllMessages work fine when constantly sending messages.
However if idle for about 15 seconds (no message in and out), both streams will stop streaming new messages.

https://github.com/xmtp/xmtp-react-native/assets/92711171/38202605-83da-46db-81f3-d0e0a1a1027f

nplasterer:
Thanks for reporting. We've looking into it. Hope to have a fix by EOD.

nplasterer:
This should be fixed in the latest version 1.29.2. Thanks again for reporting.

WilliamYWY:
Hi, I just updated the package, but I think the problem still remains.

PR #307, opened 2024-03-11
fix: Streaming bug

kele-leanes:
two tests were added to verify if the `stream` and `streamAllMessages` methods run correctly. Pointing to the local environment the tests pass whereas in the dev environment, they fail

| Local | Dev | Dev |
|:--:|:--:|:--:|
|  |  |  |


nplasterer:
Fixes: https://github.com/xmtp/xmtp-react-native/issues/306

nplasterer (approved):

PR #313, opened 2024-03-13
fix: bump android 0.8.4

kele-leanes:
empty commit for release


nplasterer:
@kele-leanes ~~I don't think this is necessary I think your PR went out yesterday with Alex release.~~ O wait looks like maybe main wasn't merged before that release.
I still think it's okay to wait on this fix till we release the streaming fix later today.

kele-leanes:
not necessary, closed

PR #318, opened 2024-03-15
Stream Validation

alexrisch:
Added tests
Added UI for hanlding streams/conversations
Ability to export events as well to recreate in normal testing pattern

https://github.com/xmtp/xmtp-react-native/assets/23103150/b7fba790-e5a2-428f-b6c5-1ae64872f6c5


nplasterer (approved):
We will probably need these for groups as well.

alexrisch:
I'll make a follow up PR based on the beta branch since there are likely to be conflicts. My only concern is that some of the tests don't necessarily pass right now but maybe a good identifier of potential issues

PR #323, opened 2024-03-19
fix: readReceipt decode method

### LINKED ISSUE #322: ReadReceiptCodec Erroring
There seems to be a small issue with read receipts content types in the RN SDK.
I see some Sentry errors `no content type found {"readReceipt":""}`
Error seems linked to this block
<https://github.com/xmtp/xmtp-react-native/blob/main/src/lib/DecodedMessage.ts#L100-L113>
From the look of it, you’re checking if `this.nativeContent[codec.contentKey]` has a value but read receipt has an empty value, as it’s supposed to

kele-leanes:
Closes #322 

Added an array `allowEmptyProperties` to accommodate keys allowing empty strings as content. For future additions, simply append new keys to this array


nplasterer (approved):
Thanks for the quick fix on this! 🙏

PR #324, opened 2024-03-22
Android Threading Fixes

nplasterer:
This fixes some technical debt we had around android threading. Going forward android should be faster and more thread safe.
All the tests pass but the complex streaming ones that we know aren't supposed to yet.
Screenshot 2024-03-21 at 5 52 25 PM


alexrisch (approved):

alexrisch:
Complex test setup likely won't work with this, we need this PR for main: https://github.com/xmtp/xmtp-react-native/pull/294
I can bring it over if you think it's worth the effort

PR #338, opened 2024-04-02
Group Chat Push Notifications

nplasterer:
Adds group chat functionality for decoding group push notifications. And updates subscribe to use new subscribe with metadata.

A big refactor of the tests and the login for the example app to address the new encryption key requirements in iOS.
Follow up with a working example here: https://github.com/xmtp/xmtp-react-native/pull/347

### DIFF
--- a/src/lib/XMTPPush.ts
+++ b/src/lib/XMTPPush.ts
@@ -1,11 +1,17 @@
+import { Client } from './Client'
 import * as XMTPModule from '../index'
 
-export class XMTPPush {
+export default class XMTPPush {
+  client: Client<any>
+
+  constructor(client: Client<any>) {
+    this.client = client
+  }
   static register(server: string, token: string) {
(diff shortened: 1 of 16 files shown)


src/lib/XMTPPush.ts
```diff
@@ -1,11 +1,17 @@
+import { Client } from './Client'
 import * as XMTPModule from '../index'
 
-export class XMTPPush {
+export default class XMTPPush {
```
cameronvoell:
```suggestion
export class XMTPPush {
```

fix build error causing failed lint CI, etc

cameronvoell: if you have a default export in XMTPPush.ts, you need to import it into index.ts before re-exporting it. If it’s a named export, you can directly re-export it in index.ts without importing it first

cameronvoell (approved):

PR #341, opened 2024-04-02
fix: Correctly handle empty/undefined fallbacks

### LINKED ISSUE #337: Bug: Fallback content is empty string when expected to be undefined
### Describe the bug

When a content type fallback is set as undefined, Decoded message's fallback is ""

### Expected behavior

Should be undefined

```
test('register and use custom content types', async () => {
  const bob = await Client.createRandom({
    env: 'local',
    codecs: [new NumberCodec()],
  })
  const alice = await Client.createRandom({
    env: 'local',
    // codecs: [new NumberCodec()],
  })
  bob.register(new NumberCodec())
  const bobConvo = await bob.conversations.newConversation(alice.address)
  const aliceConvo = await alice.conversations.newConversation(bob.address)
  await bobConvo.send(12, { contentType: ContentTypeNumber })
  const messages = await aliceConvo.messages()
  assert(messages.length === 1, 'did not get messages')
  const message = messages[0]
  let messageContent = undefined;
  try {
    messageContent = message.content()
  } catch (err) {
    messageContent = message.fallback
    console.log('Setting fallback', typeof message.fallback, {messageContent})
  }
  assert(
    messageContent !== "",
    'did not get content properly: ' + JSON.stringify(messageContent)
  )
  return true
})
```

```
class NumberCodec implements JSContentCodec<NumberRef> {
  contentType = ContentTypeNumber
  // a completely absurd way of encoding number values
  encode(content: NumberRef): EncodedContent {
    return {
      type: ContentTypeNumber,
      parameters: {
        test: 'test',
      },
      content: new TextEncoder().encode(JSON.stringify(content)),
    }
  }
  decode(encodedContent: EncodedContent): NumberRef {
    if (encodedContent.parameters.test !== 'test') {
      throw new Error(`parameters should parse ${encodedContent.parameters}`)
    }
    const contentReceived = JSON.parse(
      new TextDecoder().decode(encodedContent.content)
    ) as NumberRef
    return contentReceived
  }
  fallback(content: NumberRef): string | undefined {
    return undefined
  }
}
```

alexrisch:
Swift/Kotlin protos always set a default value for optional fields this results in differences between sdks and no real way for an undefined value to be returned even if it is expected and typed

Updated to check if there is a fallback, if there is use it, if not set back to undefined
Closes: https://github.com/xmtp/xmtp-react-native/issues/337

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/DecodedMessageWrapper.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/DecodedMessageWrapper.kt
@@ -13,14 +13,19 @@ class DecodedMessageWrapper {
             return gson.toJson(message)
         }
 
-        fun encodeMap(model: DecryptedMessage): Map<String, Any> = mapOf(
-            "id" to model.id,
-            "topic" to model.topic,
-            "contentTypeId" to model.encodedContent.type.description,
-            "content" to ContentJson(model.encodedContent).toJsonMap(),
-            "senderAddress" to model.senderAddress,
-            "sent" to model.sentAt.time,
-            "fallback" to model.encodedContent.fallback
-        )
+        fun encodeMap(model: DecryptedMessage): Map<String, Any?> {
+            // Kotlin/Java Protos don't support null values and will always put the default ""
+            // Check if there is a fallback, if there is then make it the set fallback, if not null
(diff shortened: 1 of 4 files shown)


nplasterer (approved):
Lame protos don't work but this fix seems fine. :shipit:

nplasterer:
Linking this context for future reference https://github.com/protocolbuffers/protobuf/issues/11790#issuecomment-1416495361

PR #342, opened 2024-04-03
Android Threading

### LINKED ISSUE #334: Android - Conversation List - Slow Freezes UI
The conversationList method is really slow for large conversations. It also appears to be running on the UI thread instead of a background thread and causing the UI to freeze.

nplasterer:
Closes: https://github.com/xmtp/xmtp-react-native/issues/334
This fixes the deadlocking issue for list conversations on Android because the `Coroutine` field in expo modules defaults to the Main Thread. Adding withContext for the Background thread fixes this.

### DIFF
--- a/example/src/tests.ts
+++ b/example/src/tests.ts
@@ -259,8 +259,12 @@ test('can load 1995 conversations from dev network "2k lens convos" account', as
     xmtpClient.address === '0x209fAEc92D9B072f3E03d6115002d6652ef563cd',
     'Address: ' + xmtpClient.address
   )
-
+  let start = Date.now()
   const conversations = await xmtpClient.conversations.list()
+  let end = Date.now()
+  console.log(
+    `Loaded ${conversations.length} conversations in ${end - start}ms`
+  )
   assert(
     conversations.length === 1995,
     'Conversations: ' + conversations.length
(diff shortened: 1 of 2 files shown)


example/src/tests.ts
```diff
@@ -259,8 +259,12 @@ test('can load 1995 conversations from dev network "2k lens convos" account', as
     xmtpClient.address === '0x209fAEc92D9B072f3E03d6115002d6652ef563cd',
     'Address: ' + xmtpClient.address
   )
-
+  let start = Date.now()
   const conversations = await xmtpClient.conversations.list()
+  let end = Date.now()
+  console.log(
+    `Loaded ${conversations.length} conversations in ${end - start}ms`
+  )
```
nplasterer, line 267: This is consistently giving me ~18s to load 2000 convos.

nplasterer: I was kind of thinking we could just leave this log in the tests just so we can kind of see that this number when we run it incase it gets super large etc...

cameronvoell: I like keeping this as well 👍

cameronvoell (approved):
Changes look good! Tests look good for me as well. I got 22 seconds on the 2000 convo load. Will be good to keep an eye on that for sure.

alexrisch (approved):

PR #357, opened 2024-04-11
Added documentation around sync and listing methods and parameters

cameronvoell:
It is confusing to know when a developer needs to call sync/syncGroups, and what the skipSync parameter is used for. Hopefully the documentation will help clarify. Feedback welcomed.


src/lib/Group.ts
```diff
@@ -109,6 +122,10 @@ export class Group<
     )
   }
 
+  /**
+   * Executes a network request to fetch the latest messages associated with the group
+   * and save them to the local state.
+   */
```
nplasterer, line 137: Maybe mention here as well that it also fetches the latest membership changes (since they are technically a message)

nplasterer (approved):

src/lib/Group.ts
cameronvoell: good call, added here - 20a69ad8fb6cf8365f35f6b93545d50a2ad545a6

PR #366, opened 2024-04-18
Who Added Me Feature Implementation

zombieobject:
Introduction 📟

Added the ability to identify the address of who added a member to a group.

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -804,6 +804,16 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("addedByAddress") Coroutine { clientAddress: String, id: String ->
+            withContext(Dispatchers.IO) {
+                logV("addedByAddress")
+                val client = clients[clientAddress] ?: throw XMTPException("No client")
+                val group = findGroup(clientAddress, id)
+
+                group?.addedByAddress()
+            }
+        }
+
         AsyncFunction("isGroupAdmin") Coroutine { clientAddress: String, id: String ->
             withContext(Dispatchers.IO) {
                 logV("isGroupAdmin")
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -807,10 +807,9 @@ class XMTPModule : Module() {
         AsyncFunction("addedByAddress") Coroutine { clientAddress: String, id: String ->
             withContext(Dispatchers.IO) {
                 logV("addedByAddress")
-                val client = clients[clientAddress] ?: throw XMTPException("No client")
-                val group = findGroup(clientAddress, id)
+                val group = findGroup(clientAddress, id) ?: throw XMTPException("No group found")
 
-                group?.addedByAddress()
+                group.addedByAddress()
             }
         }
 
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -705,9 +705,6 @@ public class XMTPModule: Module {
 		}
 
 		AsyncFunction("addedByAddress") { (clientAddress: String, id: String) -> String in
-			guard let client = await clientsManager.getClient(key: clientAddress) else {
-				throw Error.noClient
-			}
 			guard let group = try await findGroup(clientAddress: clientAddress, id: id) else {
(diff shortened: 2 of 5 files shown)


src/lib/Group.ts
```diff
@@ -203,6 +203,13 @@ export class Group<
     return XMTP.isGroupActive(this.client.address, this.id)
   }
 
+  async addedByAddress(skipSync = false): Promise<string> {
+    if (!skipSync) {
+      await this.sync()
+    }
+    return XMTP.addedByAddress(this.client.address, this.id)
```
alexrisch, line 207: I know this is a draft PR, just trying to add actionable steps forward:
1. The addedByAddress will need to be added in src/index.ts
2. Then it will need to be bridged over in android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt & ios/XMTPModule.swift
3. Both of which will require bumps on gradle/pod versions for your changes

zombieobject: @alexrisch Thank you so much for the advanced feedback! This will save me time and I really appreciate it. 🫶🏼

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -804,6 +804,16 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("addedByAddress") Coroutine { clientAddress: String, id: String ->
+            withContext(Dispatchers.IO) {
+                logV("addedByAddress")
+                val client = clients[clientAddress] ?: throw XMTPException("No client")
```
nplasterer: I know you were just copying the pattern in other functions but this also happens inside findGroup so is redundant. You can just remove it.
```suggestion
```

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -804,6 +804,16 @@ class XMTPModule : Module() {
...
+                val group = findGroup(clientAddress, id)
+
+                group?.addedByAddress()
```
nplasterer: We should throw if it can't find the group instead of returning null. Just like you did in iOS

ios/XMTPModule.swift
```diff
@@ -703,7 +703,18 @@ public class XMTPModule: Module {
 			
 			return try group.isActive()
 		}
-		
+
+		AsyncFunction("addedByAddress") { (clientAddress: String, id: String) -> String in
+			guard let client = await clientsManager.getClient(key: clientAddress) else {
+				throw Error.noClient
+			}
```
nplasterer: Same comment as android.
```suggestion
```

src/lib/Group.ts
```diff
@@ -203,6 +203,13 @@ export class Group<
     return XMTP.isGroupActive(this.client.address, this.id)
   }
 
+  async addedByAddress(skipSync = false): Promise<string> {
+    if (!skipSync) {
+      await this.sync()
+    }
```
nplasterer:
```suggestion
  async addedByAddress(): Promise<string> {
```

Judging from your test I actually don't think we want this sync option here at all. It's a costly call and doesn't bringing any benefit as this should be set once and not change.

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
nplasterer: Syntax wise you could probably do
```suggestion
                group?.addedByAddress() ?: throw XMTPException("No group found")
```

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
zombieobject: Will do. Appreciate the feedback.

nplasterer (approved):
Just a note if you want this to trigger a new beta release of the sdk you would want to add feat: or fix: to a commit so that it is triggered. https://github.com/semantic-release/semantic-release I'm going to be funneling through the delivery status soon so not a huge deal to leave those to release together.

zombieobject:
> Just a note if you want this to trigger a new beta release of the sdk you would want to add feat: or fix: to a commit so that it is triggered. https://github.com/semantic-release/semantic-release I'm going to be funneling through the delivery status soon so not a huge deal to leave those to release together.

Thanks for the heads up! I wasn't sure if that was pre or post-merge. The release flow is different between repos, so glad to know the process for this one. Thanks for the review. 🙂

PR #371, opened 2024-04-18
Group consent state

### LINKED ISSUE #356: Bug: group.topic incorrectly returns same value as group.id
### Describe the bug

group.topic returns the group id (same value as group.id) but messages in the group have message.topic = /xmtp/mls/1/g-<GROUP-ID>/proto , I would expect group.topic to have that same value

### Expected behavior

group.topic returns the same value as message.topic does for the messages in that group: `/xmtp/mls/1/g-<GROUP-ID>/proto`

_No response_

nplasterer:
Adds the consent state to group like we have for dms

Fixes https://github.com/xmtp/xmtp-react-native/issues/356


src/lib/Group.ts
```diff
@@ -225,4 +226,8 @@ export class Group<
       throw e
     }
   }
+
+  async consentState(): Promise<'allowed' | 'denied' | 'unknown'> {
```
peterferguson, line 227: Hey I am wondering whether this is the final state for the group `consentState`? 

I want to display which members of a group have accepted the group invite or not. AFAIK this can only be done on the client for now? 

If this function returned the state of each individual member then it would be easier to do directly from the network. 

Is this something that might be implemented in the future?

nplasterer: Hmmm not sure I understand the question? Are you looking for a way that every person in the group can see if someone else has approved or denied the group? Inherently the personal private preferences are just that private and personal. So no one else in the group can decrypt if your blocked that person or group only you can.

peterferguson: Yep exactly what I was wondering & that clarifies it.

Pending / joined members will have to just be a client-side feature in that case.

Thanks!

example/src/HomeScreen.tsx
```diff
@@ -99,12 +99,28 @@ function GroupListItem({
   const [messages, setMessages] = useState<
     DecodedMessage<SupportedContentTypes>[]
   >([])
+  const lastMessage = messages?.[0]
+  const [getConsentState, setConsentState] = useState<string | undefined>()
```
alexrisch: Normally naming conventions would be consentState

example/src/HomeScreen.tsx
```diff
@@ -99,12 +99,28 @@ function GroupListItem({
...
+
+  const denyGroup = async () => {
+    await client?.contacts.denyGroups([group.id])
+    group
+      .consentState()
+      .then(setConsentState)
```
alexrisch: This is fine to do this way, but you are already in an async function so you could also do:
```
const consent = await group
      .consentState()
      setConsentState(consent)
```

alexrisch (approved):

PR #372, opened 2024-04-19
Group message delivery status

nplasterer:
Part of https://github.com/xmtp/libxmtp/issues/516

- [x]  Add enum field for delivery status that maps to ffi
- [x]  Expose delivery status on messages
- [x]  Expose delivery status filtering on find_messages
- [x]  Expose message id on send


example/src/tests/groupTests.ts
```diff
@@ -205,11 +207,68 @@ test('production MLS V3 client creation throws error', async () => {
...
+  )
+
+  const alexMessagesFiltered: DecodedMessage[] = await alixGroup.messages(
+    true,
+    undefined,
+    undefined,
+    undefined,
+    undefined,
+    MessageDeliveryStatus.UNPUBLISHED
+  )
```
nplasterer: @alexrisch @cameronvoell this feels kind of bad for just wanting to set the delivery status? Should we be using a object like `messageOptions` instead of adding fields? This feels like how Swift and Kotlin work but doesn't seem to work out the same for typescript.

src/lib/Group.ts
```diff
@@ -110,24 +111,20 @@ export class Group<
    */
   async messages(
     skipSync: boolean = false,
```
alexrisch, line 113: Want to include skipSync in the options? You can do something like:
```
interface GroupMessagesOptions extends MessagesOptions {
skipSync?: boolean
}
```
And you can remove some duplicate code doing:
```
const {
skipSync = false
} = opts ?? {}
```

nplasterer: I'd be curious @cameronvoell thoughts here. He added skipSync all over the code base so I kind of felt like it belonged outside of the messageOptions as it's more of a conscious action. But I'm totally fine with moving it into the options if we're cool with that.

alexrisch (approved):

src/lib/Group.ts
cameronvoell: For consistency with other functions I think it's slightly better to keep it as a parameter the way it is. 

Right now feels like the optional skipSync is the least bad way I've heard for having developers be generally aware that libxmtp group functions only return results from local data, and require a sync first if you want latest from network. Open to other suggestions for sure. But if we do keep it, I think we might as well have it passed as a parameter in a consistent way across functions.

cameronvoell (approved):
looks good! left a comment with my suggestion for skipSync param

PR #395, opened 2024-05-29
feat: New V3 Identity

nplasterer:
Uses the new V3 Identity for Groups. Gets rid of accountAddress in favor of inboxId.
Will add support for smart contract wallets in a follow up PR.


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -1153,15 +1196,15 @@ class XMTPModule : Module() {
...
+        val cacheKey = "${client.inboxId}:${id}"
         val cacheGroup = groups[cacheKey]
         if (cacheGroup != null) {
             return cacheGroup
         } else {
             val group = client.conversations.listGroups()
                 .firstOrNull { it.id.toHex() == id }
             if (group != null) {
-                groups[group.cacheKey(clientAddress)] = group
+                groups[group.cacheKey(client.inboxId)] = group
```
cameronvoell, line 1207: Think this needs to be updated on line 516 as well:

https://github.com/xmtp/xmtp-react-native/blob/3542e807261f13a186be9cd68d414fa5ae1acccb/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L516

src/lib/Group.ts
```diff
@@ -197,8 +215,8 @@ export class Group<
 
   // Returns the address that added you to the group.
   // To get the latest added by address from the network, call sync() first.
```
cameronvoell, line 217: update comments here for inbox id

src/lib/Client.ts
```diff
@@ -19,16 +19,19 @@ import { DecodedMessage } from '../index'
...
 
-export type GetMessageContentTypeFromClient<C> =
-  C extends Client<infer T> ? T : never
+export type GetMessageContentTypeFromClient<C> = C extends Client<infer T>
+  ? T
+  : never
 
-export type ExtractDecodedType<C> =
-  C extends XMTPModule.ContentCodec<infer T> ? T : never
+export type ExtractDecodedType<C> = C extends XMTPModule.ContentCodec<infer T>
```
cameronvoell, line 26: can fix these lint warnings with `yarn run eslint --fix` from root proj directory

cameronvoell:
There are a bunch of tests failing because members() func was not returning properly ([issue](https://github.com/xmtp/libxmtp/issues/769)) in some cases and that was fixed in a libxmtp update yesterday in this PR - https://github.com/xmtp/libxmtp/pull/781

I'm thinking it's okay to merge this, but we should follow up with another android/ios releases to get in latest libxmtp, or we can wait with this PR, wdyt @nplasterer ?

nplasterer:
~~@cameronvoell since I used semantic commits in this one I think it would be better to wait... I just don't want a known broken package out there.~~ I think this is working okay. Lets just fast follow with the additional fixes and permissions additions.

alexrisch (approved):
Approving for fast follow with some of the updates

cameronvoell (approved):

peterferguson:
🚀

PR #402, opened 2024-06-03
Expose Group Members

### LINKED ISSUE #401: Bug: `senderAddress` returns `inboxId` in `1.34.0-beta.2`
### Describe the bug

Hey this is probably better in a discussions section but since there is a bug & no discussions I thought an issue works.

I assume this bug is just a placeholder while `inboxId` is still being implemented as it is the same in each of the client libraries.

We use `senderAddress` (and just generally the users XMTP address) extensively throughout the app. 
So I am wondering whether the longer term api will also include the `senderAddress` in messages?

It is still the case that a client is instantiated using a particular address so this could be sent with the message but maybe that is going to change? 

It is pretty essential to be able to access the available addresses related the users `inboxId` (again I assume this will happen & I am just early to the party) right now we are using a patch that just the addresses off the group

```diff
+ AsyncFunction("listMemberAddresses") { (clientAddress: String, groupId: String) -> [String: [String]] in
+	guard let group = try await findGroup(clientAddress: clientAddress, id: groupId) else {
+		throw Error.conversationNotFound("no group found for \(groupId)")
+	}
+
+	return try group.members.reduce(into: [String: [String]]()) { result, member in
+		result[member.inboxId] = member.addresses
+	}
+ }
+
```

Should I assume that the main way to get a users address will be through a method linked to the `inboxId` that will be available in the future or one of their linked addresses will be sent on the message as before?

### Expected behavior

_No response_

_No response_

nplasterer:
Part of https://github.com/xmtp/xmtp-react-native/issues/401

Exposes the members of a group and their associated addresses

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -67,6 +67,7 @@ import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import com.facebook.common.util.Hex
+import expo.modules.xmtpreactnativesdk.wrappers.MemberWrapper
 import org.xmtp.android.library.messages.MessageDeliveryStatus
 import org.xmtp.android.library.messages.Topic
 import org.xmtp.android.library.push.Service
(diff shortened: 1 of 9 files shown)


nmalzieu (approved):

peterferguson:
Super fast shipping 🚀🚀

PR #409, opened 2024-06-07
Fix group streaming

nplasterer:
WIP


nplasterer:
After increasing the delay from 100ms to 1000ms after the send the tests now pass.

Issue #410, opened 2024-06-09
Feature request: Expose Installation ID on Client


alexrisch:
### Is your feature request related to a problem?

For Groups invitations the topic is on the installation Id of the client, would like to have this exposed to the JS client

### Describe the solution to the problem

Listening for new group invitations

PR #411, opened 2024-06-10
Exposes client installationId

nplasterer:
Closes https://github.com/xmtp/xmtp-react-native/issues/410

Exposes the installationId so it is easier to construct welcome topics.


alexrisch (approved):
🚀

PR #412, opened 2024-06-10
Find Client by inboxId not address

nplasterer:
Replaces all the spots we use to store clients by address with inboxId so that we can fetch all of the content correctly going forward even after addresses are removed from an identity.


alexrisch:
Looks good code wise, however my worry is that integrators upgrading to this will have a difficult time updating as address and inboxId are both considered strings, I wonder if we should introduce using a branded type instead to help catch issues of using address/inboxId wrong

`type InboxId = string & { readonly brand: unique symbol };`

Working Typescript example:
https://www.typescriptlang.org/play/?#code/FAFwngDgpgBAkgOwEYHsAecAmMC8MDOIATgJYIDmMAZDAN4xFQCGmKCANmDEkUwpgC4YAVwQkAjsNj4wAW1TsYAXwDcwYAGM2hGCCiEAgpkyN8+XDABEABjQBGAEwBmACyXN2kLv0hEqDNh4lo6uAKyWMEzmfuhY6loIOgBmohp2FgAULCb6+EKEpBQAlLgAfHTAMFUwCfgo7FAAdOwo5FnGpvhFwEqVMB6JXikIGg6ZZP5YQjEBJTjltH1VtfVNLW0TsZjdvcDDaRl6hh253ft2hz4zWN17qQ6XhNfbdyMPRyBGOWbdQA

alexrisch (approved):

PR #419, opened 2024-06-14
Fix permissions incorrectly updating

nplasterer:
Fix inboxId regression. 

Screenshot 2024-06-13 at 6 48 01 PM


nplasterer:
So everything is passing now except `can commit after invalid permissions commit` 
I'm investigating.

cameronvoell (approved):
looks good!

PR #429, opened 2024-06-21
MLS Storage performance

nplasterer:
This bumps to a more performant mls storage. Will require everyone to clear local databases on phones.


alexrisch (approved):
Approving as iOS is coming down the line

PR #431, opened 2024-06-25
Adds custom policy type, fixes invalid policy error

cameronvoell:
Related to https://github.com/xmtp/libxmtp/pull/867

image


cameronvoell:
@alexrisch this pr should fix the invalid policy errors (tested by updating from old version and adding a new metadata field). Also "all members" now has remove member function as admin only. 

Existing groups before this update will now return "custom_policy" as their policy type since they have a remove member permission that is "allow" for all members.

nplasterer (approved):

PR #438, opened 2024-07-09
Optimistic Sendings 

nplasterer:
Adds the ability to optimistically send a messages.

Will need to call sync to actually send the message until we can figure out a way to find unpublished messages by id.

| iOS | Android |
|--------|--------|
| Screenshot 2024-07-09 at 3 48 03 PM |  Screenshot 2024-07-09 at 3 49 25 PM |


cameronvoell (approved):
looks good!

PR #439, opened 2024-07-10
Add back encryption key requirement

nplasterer:
This enforces the passing of an encryption key on create


cameronvoell (approved):
:shipit:

PR #441, opened 2024-07-16
Add a new test to debug multiple encryption keys working on iOS

nmalzieu:
Adding a new test for iOS that should pass but fails
It tries to recreate a second client with same inbox using another db encryption key
It's like the db was never encrypted to start with

NB: the thing that's supposed to throw in the try/catch does throw on Android (so that's good news!) but the try/catch doesn't seem to be able to catch it?


nplasterer:
I briefly looked into why the try catch wasn't able to catch it but couldn't figure it out. Hopefully this is just a testing issue. But let me know if you find that not to be the case.

PR #444, opened 2024-07-24
fix: remove all RunBlockings from the code

nplasterer:
This just causes potential threading issues for clients


cameronvoell (approved):
👍

PR #454, opened 2024-07-31
Add extra fields to group wrapper

### LINKED ISSUE #451: Client create method improvements
- Add a method to get the inboxId before creating the client
- Add a method to get the exact db path when the client has been created

### LINKED ISSUE #453: Add Name and URL to Group Wrapper

nplasterer:
This adds a few extra fields to the group wrapper so listing groups can have access to name and urls more efficiently.
Also exposes dbPath and a static inbox id method on client.

Closes: https://github.com/xmtp/xmtp-react-native/issues/453
Closes: https://github.com/xmtp/xmtp-react-native/issues/451

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
@@ -18,6 +18,10 @@ class GroupWrapper {
                 "topic" to group.topic,
                 "creatorInboxId" to group.creatorInboxId(),
                 "isActive" to group.isActive(),
+                "name" to group.name,
+                "imageUrlSquare" to group.imageUrlSquare,
+                "description" to group.description,
+                "pinnedFrameUrl" to group.pinnedFrameUrl
             )
         }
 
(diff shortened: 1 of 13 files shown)


android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
```diff
@@ -18,6 +18,10 @@ class GroupWrapper {
                 "topic" to group.topic,
                 "creatorInboxId" to group.creatorInboxId(),
                 "isActive" to group.isActive(),
+                "name" to group.name,
+                "imageUrlSquare" to group.imageUrlSquare,
+                "description" to group.description,
+                "pinnedFrameUrl" to group.pinnedFrameUrl
```
nplasterer, line 24: Probably no cost in adding these as well here. But I don't really see a scenario where you would need them in a group list context?

nmalzieu (approved):
This is to avoid calling the RN bridge later right?

Always a nice to have, however as discussed with Alex I think that if we managed to kill libxmtp with deadlocks, even if Converse absolutely needs to be optimized, given the # of groups we have, it means there is a bug in libxmtp we need to fix. 

LGTM

PR #458, opened 2024-08-05
Add a test for members inconsistencies when creating new installations

nmalzieu:
If you create a group with 2 inboxes, and one of those 2 inboxes creates a new installation, then sending a message on the group make it appear like a 1 person group.

This is probably causing a lot of our "group members inconsistencies"

Test fails on Android & iOS, but sometimes on iOS it also fails with  "Pool error timed out" which we know happens on iOS...

Capture d’écran 2024-08-05 à 14 33 51
Capture d’écran 2024-08-05 à 14 33 29

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -2076,12 +2076,9 @@ test('can create new installation without breaking group', async () => {
   await client1.conversations.syncGroups()
   await client2.conversations.syncGroups()
 
-  const client1Group = (await client1.conversations.listGroups()).find(
-    (g) => g.id === group.id
-  )
-  const client2Group = (await client2.conversations.listGroups()).find(
-    (g) => g.id === group.id
-  )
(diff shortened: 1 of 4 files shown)


example/src/tests/groupTests.ts
```diff
@@ -1993,6 +1994,74 @@ test('can list groups does not fork', async () => {
...
+
+  await client1.conversations.syncGroups()
+  await client2.conversations.syncGroups()
+
+  const client1Group = (await client1.conversations.listGroups()).find(
+    (g) => g.id === group.id
+  )
+  const client2Group = (await client2.conversations.listGroups()).find(
+    (g) => g.id === group.id
+  )
```
nplasterer: Total nit but you can actually call just `find_group` directly in V3 instead of iterating over the list.

cameronvoell (approved):

PR #459, opened 2024-08-05
fix: do not set pinnedframeurl when initializing group object

cameronvoell:
If the group was created before this property was added, we will throw an error. Disabling setting the pinned frame until we potentially update libxmtp to return an empty string when the field is missing.


nplasterer (approved):
You could probably just remove it entirely from that wrapper. Unless you want it there with a comment so future us knows why 😅

cameronvoell:
> You could probably just remove it entirely from that wrapper. Unless you want it there with a comment so future us knows why 😅

@nplasterer Think we might re-add it after this change in libxmtp - https://github.com/xmtp/libxmtp/issues/936

PR #462, opened 2024-08-05
Encode groups and conversations sequentially to fix iOS pool errors

cameronvoell:
image

```
 LOG  Alix loaded 50 groups in 55ms
 LOG  Alix synced 50 groups in 6ms
 LOG  Bo synced 50 groups in 89ms
 LOG  Bo loaded 50 groups in 54ms
```

### DIFF
--- a/example/src/TestScreen.tsx
+++ b/example/src/TestScreen.tsx
@@ -2,7 +2,6 @@ import { useRoute } from '@react-navigation/native'
 import React, { useEffect, useState } from 'react'
 import { View, Text, Button, ScrollView } from 'react-native'
 
-import { createdAtTests } from './tests/createdAtTests'
 import { groupPermissionsTests } from './tests/groupPermissionsTests'
 import { groupTests } from './tests/groupTests'
 import { restartStreamTests } from './tests/restartStreamsTests'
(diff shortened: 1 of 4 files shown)


ios/XMTPModule.swift
```diff
@@ -400,44 +400,31 @@ public class XMTPModule: Module {
...
-				}
-
-				var results: [String] = []
-				for try await result in group {
-					results.append(result)
-				}
-
-				return results
+			
+			var results: [String] = []
```
cameronvoell, line 404: Went ahead and removed TaskGroup from listAll and listConversations since it could lead to similar issues with too many db connections, and I did not observe any slowdown in the test for loading 2000 conversations.

before the change I saw "`LOG  Loaded 1995 conversations in 5717ms`" and after I saw actually slightly faster "`LOG  Loaded 1995 conversations in 5595ms`"

example/src/TestScreen.tsx
```diff
@@ -107,7 +106,6 @@ export enum TestCategory {
   all = 'all',
   tests = 'tests',
   group = 'group',
-  createdAt = 'createdAt',
```
alexrisch, line 110: Expected to be deleted?

nplasterer (approved):

example/src/TestScreen.tsx
cameronvoell: yea, these tests were initially just documenting different behaviors between iOS and Android, but looking at them now, they just pass / fail randomly based off random order that groups are returned. tldr, dont think these tests are useful at the moment, and are just creating extra noise, in my opinion, as the person who wrote them :-P

PR #467, opened 2024-08-09
Drop / reconnect database connections when app becomes backgrounded/active

nmalzieu:
Since we have so many issues with db locks, I suspect we have an issue calling the db disconnect at the right moment.

My thinking is that there are issues either in the Converse code or the XMTP React Native code, that blocks the bridge for too long and we don't manage to actually dicsonnect db before the app is suspended

By adding that code directly in Swift we bypass the bridge and should be able to prevent some of these crashes

Possible improvements:
- add a boolean in the config to enable / disable that feature, because it's an issue only if the db is stored in the shared container. However I don't really see how disabling that feature would bring any value
- implement this in the iOS SDK rather than the RN Module. However swift developers have access to the `Client` instance and can just call the dropDb method themselves!

Also some weirdness: why is `client.reconnectLocalDatabase` async but `client.dropLocalDatabaseConnection` sync?

### DIFF
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -62,16 +62,20 @@ public class XMTPModule: Module {
 		// A method to disconnect all dbs
 		func dropAllLocalDatabaseConnections() throws {
 			for (_, client) in clients {
-				// Call the method on each client
-				try client.dropLocalDatabaseConnection()
+				// Call the drop method on each v3 client
+                if (!client.installationID.isEmpty) {
+                    try client.dropLocalDatabaseConnection()
+                }
 			}
 		}
 
 		// A method to reconnect all dbs
 		func reconnectAllLocalDatabaseConnections() async throws {
 			for (_, client) in clients {
-				// Call the method on each client
-				try await client.reconnectLocalDatabase()
+                // Call the reconnect method on each v3 client
+                if (!client.installationID.isEmpty) {
+                    try await client.reconnectLocalDatabase()
+                }
 			}
 		}
 	}
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -48,16 +48,32 @@ public class XMTPModule: Module {
 	actor ClientsManager {
 		private var clients: [String: XMTP.Client] = [:]
 
-		// A method to update the conversations
+		// A method to update the client
 		func updateClient(key: String, client: XMTP.Client) {
 			ContentJson.initCodecs(client: client)
 			clients[key] = client
 		}
 
-		// A method to retrieve a conversation
+		// A method to retrieve a client
 		func getClient(key: String) -> XMTP.Client? {
 			return clients[key]
 		}
+        
+        // A method to disconnect all dbs
+        func dropAllLocalDatabaseConnections() throws {
+            for (_, client) in clients {
+                // Call the method on each client
+                try client.dropLocalDatabaseConnection()
+            }
+        }
+        
+        // A method to disconnect all dbs
+        func reconnectAllLocalDatabaseConnections() async throws {
+            for (_, client) in clients {
+                // Call the method on each client
+                try await client.reconnectLocalDatabase()
+            }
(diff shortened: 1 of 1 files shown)


ios/XMTPModule.swift
```diff
@@ -1459,6 +1475,18 @@ public class XMTPModule: Module {
 		  }
 		  return await client.contacts.isGroupDenied(groupId: groupId)
 		}
+
+		OnAppBecomesActive {
+			Task {
+				try await clientsManager.reconnectAllLocalDatabaseConnections()
```
neekolas: I haven't been following all the twists and turns of the database reconnection story, but have we tried doing the reconnect lazily from inside `libxmtp`? So, when we use `store.conn()` and it gets an error because the connection was dropped it automatically attempts to reconnect.

That seems less brittle than having to race to connect before we try to use the DB

nmalzieu: I'm not sure it would work. The goal is to prevent writing to db when app is backgrounded.
So if there is for instance a message that gets streamed it will not be able to write to db.
If libxmtp just reconnects the db when it wants we will keep having crashes because it might go to background, but still reopen the connection to the db when the message gets streamed, then will be killed by the os when the app gets suspended.

nmalzieu: Currently our main issue is that the db is NOT disconnected when it should

neekolas: Hmm. I see

nplasterer:
@nmalzieu you'll need a semantic commit to make this create a release. https://github.com/semantic-release/semantic-release

ios/XMTPModule.swift
```diff
@@ -47,16 +47,32 @@ public class XMTPModule: Module {
...
+				// Call the method on each client
+				try client.dropLocalDatabaseConnection()
+			}
+		}
+
+		// A method to reconnect all dbs
+		func reconnectAllLocalDatabaseConnections() async throws {
+			for (_, client) in clients {
+				// Call the method on each client
+				try await client.reconnectLocalDatabase()
```
nplasterer: This is a blocking call. Hopefully this won't create too much lag when people reopen the app if they have a lot of clients. I really think we should consider setting a limit of how many clients a user can be logged in with. For MVP I really think we should just allow 1.

nplasterer (dismissed):
Part of me thinks this change should be done on the beta branch. But I'm not sure it matters that much.

ios/XMTPModule.swift
nmalzieu: This is a blocking call but in an actor, isn't everything called in an async manner? I'm not used to actors

ios/XMTPModule.swift
```diff
@@ -1475,6 +1497,19 @@ public class XMTPModule: Module {
...
+				try await clientsManager.reconnectAllLocalDatabaseConnections()
+			}
+		}
+
+
+		OnAppEntersBackground {
+			Task {
+				try await clientsManager.dropAllLocalDatabaseConnections()
+			}
+		}
```
nplasterer, line 1510: I think we need to be a little more carful about this since these methods will error for any client that does not pass a V3Enabled flag. So they should only be called if the client is v3enabled.

ios/XMTPModule.swift
```diff
@@ -1460,6 +1480,12 @@ public class XMTPModule: Module {
 		  }
 		  return await client.contacts.isGroupDenied(groupId: groupId)
 		}
+
+		OnAppBecomesActive {
+			Task {
+				try await clientsManager.reconnectAllLocalDatabaseConnections()
+			}
+		}
```
nplasterer: Is this a duplicate of line 1504

nmalzieu: sorry about that must have been the conflict when rebasing. Fixing.

cameronvoell (approved):
Verified in example app that V2 clients seem to be working fine through multiple background and foreground actions.

PR #470, opened 2024-08-15
Fix timeout errors and add more fields to GroupWrapper

nmalzieu:
Exploring the RN thread getting completely locked sometimes, I managed to create 3 reproducible tests:

- creating 10 groups in // : not a real use case but interesting to see that creating 9 groups in // works all of the time, creating 10 groups in // fails all of the time, on iOS + Android
- listing many groups' members in // : this is closer to what we do in Converse and it's interesting to see that Android is able to list 50 groups's members in // while iOS fails at 20
- doing other stuff not linked to the database while we have locked the database => other RN methods should still work, but they get blocked until the timeout throws, the whole thread is blocked

These lock last 30 seconds (I guess that's the duration of the db lock?)
During these 30 seconds, no other call to the SDK resolves (even calls to the SDK that don't rely on the database, like `canMessage` v2 version - the call just does not reach the native code)

Also in Converse it happens in a loop because during the sync process
- try to sync => cause a 30 sec lock & timout
- after 30 sec fail => retries a sync

There is an increasing backoff for retries but it explains why it looks like the app is "bricked" forever, as soon as it gets unlocked, it goes into a lock again.

We should probably understand why these calls cause a timeout and fix it (creating 10 groups should be possible, even if we need to wait a bit for the pool, here it seems we never get back the pool connection and it times out)

We should also understand why if one async call to the sdk is locking, other calls to the sdk, even not using the database, lock as well => should a pool connection awaitance cause a thread lock?

Capture d’écran 2024-08-15 à 14 25 52

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
@@ -13,7 +13,7 @@ class GroupWrapper {
                 "clientAddress" to client.address,
                 "id" to group.id,
                 "createdAt" to group.createdAt.time,
-//                "members" to group.members().map { MemberWrapper.encode(it) },
+                "members" to group.members().map { MemberWrapper.encode(it) },
                 "version" to "GROUP",
                 "topic" to group.topic,
                 "creatorInboxId" to group.creatorInboxId(),
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
@@ -13,11 +13,12 @@ class GroupWrapper {
                 "clientAddress" to client.address,
                 "id" to group.id,
                 "createdAt" to group.createdAt.time,
-                "peerInboxIds" to group.peerInboxIds(),
+//                "members" to group.members().map { MemberWrapper.encode(it) },
                 "version" to "GROUP",
                 "topic" to group.topic,
                 "creatorInboxId" to group.creatorInboxId(),
                 "isActive" to group.isActive(),
+                "addedByInboxId" to group.addedByInboxId(),
                 "name" to group.name,
                 "imageUrlSquare" to group.imageUrlSquare,
                 "description" to group.description
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -2107,6 +2107,50 @@ test('can create new installation without breaking group', async () => {
   return true
 })
 
+test('can create 10 groups in parallel', async () => {
+  const [client1, client2] = await createClients(2)
+  const groupsPromise: Promise<Group>[] = []
+
+  // Creating 9 groups in // works
+  for (let index = 0; index < 9; index++) {
+    groupsPromise.push(client1.conversations.newGroup([client2.address]))
+  }
+  console.log('Creating 9 groups...')
+  await Promise.race([
+    Promise.all(groupsPromise),
+    new Promise((_, reject) =>
+      setTimeout(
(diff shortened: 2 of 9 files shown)


nplasterer:
This PR now Fixes https://github.com/ephemeraHQ/converse-app/issues/422 and Fixes https://github.com/xmtp/libxmtp/issues/955
And adds functionality for `addedBy` and `members` list to the group

Closes https://github.com/xmtp/xmtp-react-native/issues/476
Closes https://github.com/xmtp/xmtp-react-native/issues/475

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/GroupWrapper.kt
```diff
@@ -13,11 +13,12 @@ class GroupWrapper {
                 "clientAddress" to client.address,
                 "id" to group.id,
                 "createdAt" to group.createdAt.time,
-                "peerInboxIds" to group.peerInboxIds(),
+                "members" to group.members().map { MemberWrapper.encode(it) },
```
nplasterer, line 16: I didn't want to duplicate the DB call to members so this is a breaking change for anyone who is using `peerInboxIds` directly but is easily mitigated by mapping on inboxID if desired.

example/src/tests/groupTests.ts
```diff
@@ -2100,13 +2085,32 @@ test('can create new installation without breaking group', async () => {
...
+  }
+
+  try {
+    await Promise.all(groups.slice(0, 20).map((g) => g.membersList()))
+  } catch (e) {
+    throw new Error(`Failed listing 20 groups members with ${e}`)
+  }
+
+  return true
+})
```
nplasterer, line 2111: Only left this one test that was previously failing for me and passing now.

src/lib/Group.ts
```diff
@@ -634,7 +631,7 @@ export class Group<
    * @returns {Promise<Member[]>} A Promise that resolves to an array of Member objects.
    * To get the latest member list from the network, call sync() first.
    */
-  async members(): Promise<Member[]> {
+  async membersList(): Promise<Member[]> {
     return await XMTP.listGroupMembers(this.client.inboxId, this.id)
   }
```
nplasterer, line 637: Instead of removing this method renamed it. Since typescript wouldn't allow the name collision. It's maybe not necessary?

example/src/tests/groupTests.ts
```diff
@@ -2047,12 +2037,9 @@ test('can create new installation without breaking group', async () => {
...
     166, 83, 208, 224, 254, 44, 205, 227, 175, 49, 234, 129, 74, 252, 135, 145,
   ])
-  const wallet1 = new Wallet(
-    '0xc54c62dd3ad018ef94f20f0722cae33919e65270ad74f2d1794291088800f788'
-  )
-  const wallet2 = new Wallet(
-    '0x8d40c1c40473975cc6bbdc0465e70cc2e98f45f3c3474ca9b809caa9c4f53c0b'
-  )
+  const wallet1 = Wallet.createRandom()
+  const wallet2 = Wallet.createRandom()
```
nplasterer, line 2041: These wallets are in the bad [key package](https://github.com/xmtp/libxmtp/issues/947) state so had to use random ones. I think this is better anyways. But good to know for future testing when we have a fix.

nplasterer (approved):

nplasterer:
| Android | iOS |
|--------|--------|
|Screenshot 2024-08-19 at 4 39 40 PM |Screenshot 2024-08-19 at 5 06 31 PM |

src/lib/Group.ts
nmalzieu: @nplasterer the method is still useful
If we call group.sync() , I don't think the value of the attribute group.members will be updated in place? So being able to call group.membersList() to get the refreshed value is useful?

PR #482, opened 2024-08-26
fix: syncAllGroups returns num of groups synced. possible fork fixes

cameronvoell:


cameronvoell:
Interesting that for some reason on android RN, the first syncAllGroups after being removed reflects the new number of groups the user is in, while on android native, ios native, and ios RN, one sync is required after removal from a group, before it to be reflected in the numActive groups. 

This inconsistency should not affect any core behavior and isnt worth holding up the rest of the fixes and updates in this release, so I recommend we move ahead and look into that behavior inconsistency in a different thread.

See: https://github.com/xmtp/xmtp-react-native/pull/482/files#diff-3b2eca62a1b7bff3da3b3a36c6afedb29d404f511354c40bcf6513e631bd0732R2177-R2190

neekolas (approved):

alexrisch (approved):

PR #487, opened 2024-08-30
Canceling streams cancels underlying stream

nplasterer:
Confirm that when we call stream cancel that the underlying stream is actually canceled.

| iOS | Android |
|--------|--------|
| Screenshot 2024-09-02 at 9 44 12 PM | Screenshot 2024-09-02 at 9 48 52 PM |

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -42,929 +42,924 @@ async function createGroups(
   return groups
 }
 
-test('can make a MLS V3 client', async () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const keyBytes = new Uint8Array([
-    233, 120, 198, 96, 154, 65, 132, 17, 132, 96, 250, 40, 103, 35, 125, 64,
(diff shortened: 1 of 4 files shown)


nmalzieu (approved):

example/src/tests/groupTests.ts
```diff
@@ -957,6 +958,52 @@ test('can remove and add members from a group by inbox id', async () => {
...
+  await delayToPropogate()
+
+  assert(
+    messageCallbacks === 2,
+    'message stream should have received 2 message'
+  )
+
+  return true
+})
+
```
nmalzieu, line 1006: @nplasterer  was this test failing before? I thought streamAllMessages WAS canceled even if underlying stream was not and so this test would have passed before?

nmalzieu (approved):

example/src/tests/groupTests.ts
nplasterer: Yes the test passed before as well. But this shows the logs easier which is how I was debugging.

PR #491, opened 2024-09-12
FAILING test in dev env for streaming groups and messages at the same time

nmalzieu:
While working on group invite links, I noticed that when joining a group, sometimes Converse didn't detect the new group that had just been joined and did not redirect to it
By commenting parts of the code it seems to be linked to streaming groups & messages at the same time
I actually can make the attached test fail in `dev` env but not in `local`: if I streamAllMessages, I can't stream groups anymore!

Capture d’écran 2024-09-12 à 15 59 53

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -1248,29 +1248,23 @@ test('can stream all groups and conversations', async () => {
 })
 
 test('can stream groups and messages', async () => {
-  for (const env of ['local', 'dev'] as const) {
-    const [alixClient, boClient] = await createClients(2, env)
+  const [alixClient, boClient] = await createClients(2)
 
-    // Start streaming groups
-    const groups: Group<any>[] = []
-    await alixClient.conversations.streamGroups(async (group: Group<any>) => {
-      groups.push(group)
-    })
-    // Also stream messages
-    await alixClient.conversations.streamAllMessages(
-      async (message) => {},
-      true
-    )
+  // Start streaming groups
+  const groups: Group<any>[] = []
+  await alixClient.conversations.streamGroups(async (group: Group<any>) => {
+    groups.push(group)
+  })
+  // Stream messages twice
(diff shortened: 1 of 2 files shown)


example/src/tests/groupTests.ts
```diff
@@ -1247,6 +1247,35 @@ test('can stream all groups and conversations', async () => {
   return true
 })
 
+test('can stream groups and messages', async () => {
+  for (const env of ['local', 'dev'] as const) {
```
nmalzieu: it would be the only test in that test suite that runs not only on `local` env. Is that ok?

example/src/tests/groupTests.ts
```diff
@@ -1247,6 +1247,35 @@ test('can stream all groups and conversations', async () => {
...
+    // Also stream messages
+    await alixClient.conversations.streamAllMessages(
+      async (message) => {},
+      true
+    )
+
+    // bo creates a group with alix so a stream callback is fired
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    await boClient.conversations.newGroup([alixClient.address])
+    await delayToPropogate()
```
cameronvoell: I tested passing in a longer value to delayToPropogate here (default is 100 ms). Passing in 2000 ms, test passed 5/5 times for me on ios and android. There could just be network delays initiating the stream. 

To better deal with this flakiness, I think we need to update these tests to use Promise/Resolve instead of relying on delays, see this draft PR: https://github.com/xmtp/xmtp-react-native/pull/284/files

(I think at the time that PR was made, there was other issues that were breaking streams, but now that streams are more stable now might be a good time to revisit that improvement for tests)

nmalzieu: ah interesting, so I might have written a wrong test for a real problem I have in the app (that has nothing to do with 100ms I was waiting like 10 secs). I'll keep investigating then.

nplasterer:
I think this is fixed now.

PR #492, opened 2024-09-13
Bump the versions for plaintext headers

nplasterer:
Bump to the latest version of the sdk to get the plaintext headers


cameronvoell (approved):
don't forget semantic release comment 👍

PR #496, opened 2024-09-19
Bug streaming groups multi client

### LINKED ISSUE #495: 2 Clients logged in on the same app streaming all messages
Group streams seem to receive all message for all clients not just the one that can decrypt it.

nplasterer:
Fixes https://github.com/xmtp/xmtp-react-native/issues/495

### DIFF
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -1262,7 +1262,7 @@ test('can stream groups and messages', async () => {
   // bo creates a group with alix so a stream callback is fired
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   await boClient.conversations.newGroup([alixClient.address])
-  await delayToPropogate(2000)
+  await delayToPropogate()
   if ((groups.length as number) !== 1) {
     throw Error(`Unexpected num groups (should be 1): ${groups.length}`)
   }
--- a/example/src/tests/groupTests.ts
+++ b/example/src/tests/groupTests.ts
@@ -43,6 +43,2247 @@ async function createGroups(
   return groups
 }
 
+test('can make a MLS V3 client', async () => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const keyBytes = new Uint8Array([
+    233, 120, 198, 96, 154, 65, 132, 17, 132, 96, 250, 40, 103, 35, 125, 64,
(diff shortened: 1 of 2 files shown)


nplasterer:
Does not reproduce in Android proper. Only reproduces in iOS and Android inside of ReactNative. Makes me think it's an event listener issue.

nplasterer:
All the other methods already had this check so everything should work now. 🙏

example/src/tests/groupTests.ts
```diff
@@ -1247,6 +1247,29 @@ test('can stream all groups and conversations', async () => {
...
+  // bo creates a group with alix so a stream callback is fired
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  await boClient.conversations.newGroup([alixClient.address])
+  await delayToPropogate(2000)
+  if ((groups.length as number) !== 1) {
+    throw Error(`Unexpected num groups (should be 1): ${groups.length}`)
+  }
+
+  return true
+})
```
nplasterer, line 1271: @nmalzieu I think this fix also fixed this weird bug? Can you confirm?

nplasterer:
Screenshot 2024-09-19 at 10 37 10 PM

src/lib/Conversations.ts
```diff
@@ -146,7 +146,7 @@ export default class Conversations<
     const groupsSubscription = XMTPModule.emitter.addListener(
       EventTypes.Group,
       async ({ inboxId, group }: { inboxId: string; group: GroupParams }) => {
-        if (this.known[group.id]) {
+        if (this.known[group.id] || this.client.inboxId !== inboxId) {
```
cameronvoell, line 149: I think this is a good fix to merge now. 

long term, I wonder if it would be a worthwhile performance gain to programmatically generate these Expo Module events to include the inbox-id in their name, so we dont send copies of all clients' events to each client event listener and then need to filter them out. 
https://github.com/xmtp/xmtp-react-native/blob/08341824ca289797405d8494eb553b74113f1b2c/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L214-L231

cameronvoell (approved):
nice fix! 🚢

src/lib/Conversations.ts
nplasterer: I looked into that and it didn't seem straight forward. But something we should consider looking into again for sure.

PR #503, opened 2024-09-26
Noe/fix group streams issues

nmalzieu:
- removed the check for `this.known` when streaming a group because we might call somewhere else `listGroups` which will mark the group as known but we still want it to get streamed. If there are other effects of this check @nplasterer let's discuss!
- added a small delay in a test after streaming to be sure the stream is opened before sending a message
- added a failing test when streaming both messages & groups in iOS, that is fixed by this Swift PR: https://github.com/xmtp/xmtp-ios/pull/406

### DIFF
--- a/src/lib/Conversations.ts
+++ b/src/lib/Conversations.ts
@@ -146,7 +146,7 @@ export default class Conversations<
     const groupsSubscription = XMTPModule.emitter.addListener(
       EventTypes.Group,
       async ({ inboxId, group }: { inboxId: string; group: GroupParams }) => {
-        if (this.known[group.id] || this.client.inboxId !== inboxId) {
+        if (this.client.inboxId !== inboxId) {
           return
         }
         this.known[group.id] = true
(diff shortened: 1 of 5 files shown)


src/lib/Conversations.ts
```diff
@@ -146,7 +146,7 @@ export default class Conversations<
     const groupsSubscription = XMTPModule.emitter.addListener(
       EventTypes.Group,
       async ({ inboxId, group }: { inboxId: string; group: GroupParams }) => {
-        if (this.known[group.id] || this.client.inboxId !== inboxId) {
```
nplasterer, line 149: The forked group test is still flaky for me so I don't think this fixes it. I'd probably lean more towards reverting this change just in case. But also okay with leaving it if you think it will fix other issues as all the tests seem to be passing.

nplasterer (approved):
Don't forget to add a [semantic commit so a release ](https://github.com/semantic-release/semantic-release) gets triggered.

PR #508, opened 2024-10-02
Smart contract sdk support

nplasterer:
Pulled and cleaned up the SDK pieces from https://github.com/xmtp/xmtp-react-native/pull/501


ios/ReactNativeSigner.swift
```diff
@@ -14,11 +14,18 @@ class ReactNativeSigner: NSObject, XMTP.SigningKey {
...
 	var continuations: [String: CheckedContinuation<XMTP.Signature, Swift.Error>] = [:]
+	var scwContinuations: [String: CheckedContinuation<Data, Swift.Error>] = [:]
 
-	init(module: XMTPModule, address: String) {
+	init(module: XMTPModule, address: String, walletType: WalletType = WalletType.EOA, chainId: Int64? = nil, blockNumber: Int64? = nil) {
 		self.module = module
 		self.address = address
+		self.walletType = walletType
+		self.chainId = chainId
+		self.blockNumber = blockNumber
```
neekolas, line 28: I don't think we need the block number to be stored on the signer. That should get resolved at the time of signing, not on instantiation, since the block number keeps incrementing.

src/index.ts
```diff
@@ -220,7 +225,10 @@ export async function createOrBuild(
   enableV3?: boolean | undefined,
   dbEncryptionKey?: Uint8Array | undefined,
   dbDirectory?: string | undefined,
-  historySyncUrl?: string | undefined
+  historySyncUrl?: string | undefined,
+  walletType?: WalletType | undefined,
+  chainId?: number | undefined,
+  blockNumber?: number | undefined
```
neekolas, line 231: Don't think we need the block number here either. This only matters when you are actually signing something.

ios/ReactNativeSigner.swift
nplasterer: We store it on the `SigningKey` so it only gets passed in from the `SigningKey` which means we only have it when the signing key is present which is only when something is being signed. Unfortunately with the bridge this is how it has to be passed. But this and the below shouldn't be exposed to users so it should be an internal matter.

neekolas: OK. I can approve the PR. Still feels wrong to get that block number even one second before we actually sign the message, since it's going to be older than the block the signature was actually generated on.

neekolas (approved):

ios/ReactNativeSigner.swift
nplasterer: Hmmm okay maybe we need to rethink how I have this setup inside the other sdks as well. 🤔 

As they all work the same with the block number as part of the signingKey https://github.com/xmtp/xmtp-android/blob/main/library/src/main/java/org/xmtp/android/library/SigningKey.kt#L31-L34 But maybe it should be passed as a param to create instead?

neekolas: Ideally it would get pulled from the chain at the same time you are signing the message.

neekolas: Another way of thinking about it would be to have a `getBlockNumber` method that is only implemented on SCWs, which we could call from our SDKs when we are gathering a smart contract signature.

PR #514, opened 2024-10-21
Return last message if passed as group option

### LINKED ISSUE #513: Bug: Last Message is not exposed in Group type when last message is included in list groups
### Describe the bug

In Group.ts lastMessage is not being exposed when returned from the listGroups method

### Expected behavior

Last message is exposed and decrypted

_No response_

nplasterer:
Closes https://github.com/xmtp/xmtp-react-native/issues/513

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -635,7 +635,7 @@ class XMTPModule : Module() {
                 val sortedGroupList = if (order == ConversationOrder.LAST_MESSAGE) {
                     client.conversations.listGroups()
                         .sortedByDescending { group ->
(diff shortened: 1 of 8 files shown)


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -635,7 +635,7 @@ class XMTPModule : Module() {
                 val sortedGroupList = if (order == ConversationOrder.LAST_MESSAGE) {
                     client.conversations.listGroups()
                         .sortedByDescending { group ->
-                            group.decryptedMessages(limit = 1).firstOrNull()?.sentAt
```
nplasterer, line 638: I actually ran into this issue on the native side. Limit actually just picks the first message where as we want the last message. So we can't use limit here. Could be an improvement in libxmtp to allow the order to be passed as a parameter.

alexrisch (approved):

PR #526, opened 2024-11-08
Update to a V3 only SDK

nplasterer:
Remove V1 & V2 Conversation interfaces
Remove unneeded decrypt method 
Remove contacts and replace with private preferences 
Remove V2 client and associated classes 
Big Overhaul of the tests


nplasterer:
All the tests are green! 🎉 

| Android | iOS |
|--------|--------|
| Screenshot 2024-11-11 at 8 37 11 PM  |  Screenshot 2024-11-11 at 8 51 05 PM | 

Some follow ups that would be nice:
- Polish up the example app
- Updating the send messages to mirror how v2 convos handled it
- Take pass on performance + tests
- Frames client
- Any additional polish

example/src/tests/createdAtTests.ts
```diff
@@ -213,7 +211,7 @@ test('group createdAt matches streamAll', async () => {
     'second ' + allGroups[1].createdAt + ' != ' + caroGroup.createdAt
   )
 
-  cancelStream()
+  alix.conversations.cancelStream()
```
codabrink, line 212: Do we want to revert this line back to what it was or delete the above `cancelStream()`?

codabrink (approved):
This all looks really good, nice work!

example/src/tests/createdAtTests.ts
nplasterer: No I removed the way cancel streams works for conversations. So you have to call cancel stream directly now.

PR #538, opened 2024-11-18
Making fork repro test smaller

cameronvoell:
```typescript
test('groups cannot fork', async () => {
  const [alix, bo, new_one, new_two] = await createClients(4)
  // Create group with 2 users
  const alixGroup = await alix.conversations.newGroup([
    bo.address,
    new_one.address,
    new_two.address,
  ])

  // sync clients
  await alix.conversations.sync()
  await bo.conversations.sync()
  const boGroup: Group<DefaultContentTypes> = (await bo.conversations.findGroup(alixGroup.id))!

  // Remove two members in parallel
  // NB => if we don't use Promise.all but a loop, we don't get a fork
  console.log('*************libxmtp*********************: Removing members in parallel')
  await Promise.all([
    alixGroup.removeMembers([new_one.address]),
    alixGroup.removeMembers([new_two.address])
  ])
```

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -98,19 +98,19 @@ repositories {
 dependencies {
   implementation project(':expo-modules-core')
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${getKotlinVersion()}"
-  implementation "org.xmtp:android:3.0.5"
+  // implementation "org.xmtp:android:3.0.5"
   implementation 'com.google.code.gson:gson:2.10.1'
   implementation 'com.facebook.react:react-native:0.71.3'
   implementation "com.daveanthonythomas.moshipack:moshipack:1.0.1"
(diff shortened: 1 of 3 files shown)


cameronvoell:
Test code here is causing a fork due to client processing their own message as an external message. Database locked is showing in the logs, likely cause

https://github.com/xmtp/xmtp-react-native/blob/8a1c4883a35803e68a4407212eb66636c44d292d/example/src/tests/groupTests.ts#L1641-L1663

neekolas:
A variation on this we'll want to test as well is what happens if the two removes happen in parallel, but are being done by different group members. That can happen in the real world as well (me and you both kick someone out of the group at the same time right after an installation was revoked)

cameronvoell:
Fixed in https://github.com/xmtp/xmtp-react-native/pull/569, test added to groupTests file

PR #540, opened 2024-11-19
Add ability to sign with an installation key

nplasterer:
This adds the ability to sign with an installation key
Also fixes a TLS issue with android.

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
@@ -364,6 +364,15 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("signWithInstallationKey") Coroutine { inboxId: String, message: String ->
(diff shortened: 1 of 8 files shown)


insipx (approved):

android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -364,6 +364,15 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("signWithInstallationKey") Coroutine { inboxId: String, message: String ->
+            withContext(Dispatchers.IO) {
+                val client = clients[inboxId] ?: throw XMTPException("No client")
+
+                val signature = client.signWithInstallationKey(message)
+                signature.map { it.toInt() and 0xFF }
```
mchenani, line 372: Suggestion: can we add a comment or extract this value?

mchenani (approved):
Looks good ✌️, also added one comment🙌

PR #542, opened 2024-11-20
processMessage method

nmalzieu:
the processConversationMessage method needs an `await` and the result is already an object not a json string

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1450,12 +1450,12 @@ export async function processConversationMessage<
   id: string,
   encryptedMessage: string
 ): Promise<DecodedMessage<ContentTypes>> {
-  const json = await XMTPModule.processConversationMessage(
+  const obj = await XMTPModule.processConversationMessage(
     client.inboxId,
     id,
     encryptedMessage
   )
-  return DecodedMessage.fromObject(json, client)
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1450,12 +1450,12 @@ export async function processConversationMessage<
...
   encryptedMessage: string
 ): Promise<DecodedMessage<ContentTypes>> {
-  const json = XMTPModule.processConversationMessage(
+  const obj = await XMTPModule.processConversationMessage(
     client.inboxId,
     id,
     encryptedMessage
   )
-  return DecodedMessage.from(json, client)
+  return DecodedMessage.fromObject(obj, client)
```
nplasterer:
```suggestion
  return DecodedMessage.from(obj, client)
```

nplasterer:
To be able to do a legacy release you'll need to bump this version number https://github.com/xmtp/xmtp-react-native/blob/xmtp-legacy/package.json#L3

Last legacy release was [2.8.4](https://www.npmjs.com/package/@xmtp/react-native-sdk/v/2.8.4) So I would bump this to `2.8.5`

nplasterer (approved):

PR #543, opened 2024-11-20
fix(processMessage): await processMessage and handle object

nmalzieu:
Same as https://github.com/xmtp/xmtp-react-native/pull/542 but not legacy

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -857,12 +857,12 @@ export async function processMessage<
   id: ConversationId,
   encryptedMessage: string
 ): Promise<DecodedMessage<ContentTypes>> {
-  const obj = await XMTPModule.processMessage(
+  const json = await XMTPModule.processMessage(
     client.inboxId,
     id,
     encryptedMessage
   )
-  return DecodedMessage.fromObject(obj, client)
+  return DecodedMessage.from(json, client)
 }
 
 export async function processWelcomeMessage<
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -857,8 +857,12 @@ export async function processMessage<
...
   encryptedMessage: string
 ): Promise<DecodedMessage<ContentTypes>> {
-  const json = XMTPModule.processMessage(client.inboxId, id, encryptedMessage)
-  return DecodedMessage.from(json, client)
+  const obj = await XMTPModule.processMessage(
+    client.inboxId,
+    id,
+    encryptedMessage
+  )
+  return DecodedMessage.fromObject(obj, client)
```
nplasterer:
```suggestion
  return DecodedMessage.from(json, client)
```

nplasterer: The thing returned from XmtpModule is json not a object you have to do from here like we do in other places in the index.

https://github.com/xmtp/xmtp-react-native/blob/a955447c1eff205418be567e75a0369859d3bc7c/src/index.ts#L351-L359

nmalzieu: This was failing with json parse error while my code was working! Logging the json variable showed it was an object not a string?

nplasterer: It 100% returns a string https://github.com/xmtp/xmtp-react-native/blob/main/ios/XMTPModule.swift#L1309.

nplasterer (approved):

PR #544, opened 2024-11-22
Mutli Wallet Support

nplasterer:
Add the ability to add and remove accounts to an existing inboxId

Supports SCW as well


android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt
```diff
@@ -373,6 +421,17 @@ class XMTPModule : Module() {
             }
         }
 
+        AsyncFunction("verifySignature") Coroutine { inboxId: String, message: String, signature: List<Int> ->
+            withContext(Dispatchers.IO) {
```
mchenani, line 426: Suggestion: I see in all other functions we the name of the function. Can we have it here too?
```
logV("verifySignature")
```

mchenani (approved):

src/lib/Group.ts
```diff
@@ -281,7 +281,7 @@ export class Group<
   /**
    * Returns the group image url square.
    * To get the latest group image url square from the network, call sync() first.
-   * @returns {string} A Promise that resolves to the group name.
+   * @returns {string} A Promise that resolves to the group image url.
```
cameronvoell, line 284: 🙏

cameronvoell (approved):

PR #546, opened 2024-11-25
Add ability to stream consent

nplasterer:
Adds the ability to stream consent records
Small rename to make consent records more accurate


src/lib/types/EventTypes.ts
```diff
@@ -17,4 +17,8 @@ export enum EventTypes {
    * A new message is sent to a specific conversation
    */
   ConversationMessage = 'conversationMessage',
+  /**
+   * A inboxId or conversation has been approved or denied
```
mchenani, line 21: Question: is that a correct comment?

mchenani (approved):
🙌, added one question.

src/lib/types/EventTypes.ts
nplasterer: Yes you get a consent stream whenever an inboxId or conversation has been approved or denied.

PR #547, opened 2024-11-26
History Sync tests

nplasterer:
Adds the ability to test history sync by having multiple installations with the same inboxId running at the same time.

In React Native we keep a map of clients by inboxId in the bridge so that you can easily fetch a client and call it's respective functions on it. This makes it so you can only have one client at a time. Unfortunately when it comes to installations we want a client per installation the two options are to 

1) change our client mapping to be by installationId
2) add very specific tests for history syncing that give access to these multiple installations

My gut is that doing this based on inboxId is going to bite us in the future and that it should really be mapped by installationId. However thats going to be a massive change.


nplasterer:
Closing in favor of option 1

PR #548, opened 2024-11-26
Make the bridge installationId based

nplasterer:
Adds the ability to test history sync by having multiple installations with the same inboxId running at the same time.

In React Native we keep a map of clients by inboxId in the bridge so that you can easily fetch a client and call it's respective functions on it. This makes it so you can only have one client at a time. Unfortunately when it comes to installations we want a client per installation the two options are to

This changes our client mapping to be by installationId


nplasterer:
All tests are green!

mchenani (approved):
🧹

cameronvoell (approved):
updates and tests all looking good 👍

PR #550, opened 2024-11-27
Sync all syncs all

### LINKED ISSUE #555: Static Can Message
Bring back the static Can message

nplasterer:
- Updates the sync all conversations to actually sync all
- Updates tests for performance
- Adds dependency on SQL cipher.
- Small performance improvement
- Adds static can message
Closes https://github.com/xmtp/xmtp-react-native/issues/555


nplasterer:
Consistently getting `Built a client with inboxId in 369.93345899973065ms` across all sdks and RN. I think it's fair to say that this should hopefully be the experience across the board.

Client create is also marginally better but we still have a lot of room to improve here.

codabrink (approved):

PR #552, opened 2024-11-28
feat: Decoded Message Types

alexrisch:
Please enjoy this [Thanksgiving Gif](https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExeGVtM2hvNGZwaTMzc3pybWU0N3FyNndhNmR5ZmE3dmxvaml0OXB4YiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/dWknHx8LmjvqCwdK1U/giphy.gif)

Improved Decoded Message types and returns from listing 
Added DecodedMessageUnion utility type to manage returned values from list methods
Added more type tests to help check against type usage

One callout is to really make types work I am using as const for the codecs, I did this rather than requiring readonly array of codecs. This means integrators can still dynamically create codecs but still use types, but this is only needed for the initial type inferencing if done by the client

### DIFF
--- a/example/src/types/typeTests.ts
+++ b/example/src/types/typeTests.ts
@@ -2,6 +2,7 @@ import {
   Client,
   ContentTypeId,
   ConversationVersion,
+  DecodedMessage,
   Dm,
   EncodedContent,
   JSContentCodec,
@@ -9,6 +10,7 @@ import {
   ReplyCodec,
   sendMessage,
   TextCodec,
+  DecodedMessageUnion,
 } from 'xmtp-react-native-sdk'
 
 const ContentTypeNumber: ContentTypeId = {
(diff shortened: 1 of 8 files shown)


example/src/types/typeTests.ts
```diff
@@ -169,8 +171,8 @@ export const typeTests = async () => {
       topNumber: {
         bottomNumber: 12,
       },
-    },
-    { contentType: ContentTypeNumber }
+    }
+    // { contentType: ContentTypeNumber }
```
alexrisch, line 175: These were already failing, V3 related I believe

src/lib/DecodedMessage.ts
```diff
@@ -21,6 +21,7 @@ export enum MessageDeliveryStatus {
 }
 
 export class DecodedMessage<
+  ContentType extends DefaultContentTypes[number] = DefaultContentTypes[number],
```
thierryskoda, line 24: Maybe we could have a type for `DefaultContentTypes[number]` like 
`export type DefaultContentType = DefaultContentTypes[number]`

since we use it a lot

thierryskoda (approved):

PR #85, opened 2023-08-16
feat: support remote attachments

dmccartney:
This adds support for remote attachments to the RN SDK.

Goals: do heavy lifting in native code and minimize data sent across JS/native bridge (i.e. use file references).

Note: this is the binding for `remoteAttachment` leftover from #81 and uses a similar approach to abstracting codec usage.

Sending:
- the app helps the sender select a `fileUri` to send
- the app asks XMTP to encrypt the file: `client.encryptAttachment({ fileUri, ... })`
- the app uploads the `encryptedLocalFileUri` to web storage to get a `url`
- the app asks XMTP to send the message: `client.send({ remoteAttachment: { url, ... }})`

Screenshot 2023-08-16 at 3 55 27 PM
Screenshot 2023-08-16 at 3 55 40 PM

Receiving:
- the app lists messages from XMTP having a `message.content.remoteAttachment`
- when the receiver taps the attachment, the app fetches the `url` to an `encryptedLocalFileUri`
- the app asks XMTP to decrypt the local file: `client.decryptAttachment({ encryptedLocalFileUri, ...})`
- the app presents the decrypted `fileUri` to the user based on the `mimeType`

Screenshot 2023-08-16 at 3 55 50 PM
Screenshot 2023-08-16 at 3 55 58 PM

This is the high-level TODO list:

- [x] sending
  - [x] JS interface + demo sending UI
  - [x] iOS sending
  - [x] Android sending
- [x] receiving
  - [x] JS interface + demo receiving UI
  - [x] iOS receiving
  - [x] Android receiving
- misc
  - [ ] cleanup the `ConversationScreen` code in the demo app
  - [ ] find a less hacky dev rigging for `upload-service` testing

PR #558, opened 2024-12-07
Fixes remote attachments testing in example app

cameronvoell:
Revives and exposes commands for remote attachment testing from previous work https://github.com/xmtp/xmtp-react-native/pull/85

See new README instructions:

> If testing the remote attachment content type, run (from example directory):
```bash
yarn run upload:up
```
> If running on andoid also run:
```bash
adb reverse tcp:8443 tcp:8443
```

image


nplasterer (approved):
What do you think about bringing this test back from the legacy branch as well?

https://github.com/xmtp/xmtp-react-native/blob/xmtp-legacy/example/src/tests/tests.ts#L686-L773

I think a few tests from that file are probably still relevant even though I removed the whole file in the upgrade.

PR #559, opened 2024-12-07
Add back custom content type ability.

nplasterer:
Adds back the ability to add custom content types in V3.


cameronvoell (approved):
tests passing, just needed minor tweak on ios/android libs 👍

PR #582, opened 2025-01-03
V3 HMAC key support for self push notifications

nplasterer:
Adds ability to get hmac keys for conversations to help with self push notifications
Also adds a stream to know when to refetch hmac keys

### DIFF
--- a/ios/XMTPModule.swift
+++ b/ios/XMTPModule.swift
@@ -104,7 +104,8 @@ public class XMTPModule: Module {
 			"conversation",
 			"message",
 			"conversationMessage",
-			"consent"
+			"consent",
+			"preferences"
 		)
 
 		AsyncFunction("address") { (installationId: String) -> String in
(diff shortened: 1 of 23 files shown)


nplasterer:
Screenshot 2025-01-08 at 8 01 38 AM

ios/XMTPModule.swift
```diff
@@ -582,6 +578,17 @@ public class XMTPModule: Module {
 			return results
 		}
 
+		AsyncFunction("getHmacKeys") { (installationId: String) -> [UInt8] in
+			guard
+				let client = await clientsManager.getClient(key: installationId)
+			else {
+				throw Error.noClient
+			}
+			let hmacKeys = try await client.conversations.getHmacKeys()
```
codabrink, line 587: No need for async here

codabrink (approved):
Other than the one comment, I think it looks good!

ios/XMTPModule.swift
nplasterer: So getHmacKeys is not async but because conversations is an actor for thread safety it has to be be awaited.

PR #595, opened 2025-01-30
build(deps): patch react-native for boost issue

MuckT:
### Description

I got an error when I tried to run locally related to boost: https://github.com/boostorg/boost/issues/843, this should fix that error.


cameronvoell:
@MuckT Thanks for flagging this! I ended up fixing it using these instruction posted on the [facebook/react-native](https://github.com/facebook/react-native/issues/42180) repo, which had a slightly different archive URL to resolve the checksum error. I'll go ahead and close this PR, since it is now resolved on the `main` branch. But you saved us some time finding the solution so thanks!

https://github.com/xmtp/xmtp-react-native/pull/596#discussion_r1939891814

PR #596, opened 2025-01-30
Messages with reactions function, dm fix

cameronvoell:
To do:
- [x] update android ref when https://github.com/xmtp/xmtp-android/pull/363 is merged and released
- [x] update ios ref and implement ios side
- [x] fix release action so we create a tag when releasing


android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
```diff
@@ -159,6 +177,18 @@ class ContentJson(
                 )
             )
 
+            ContentTypeReactionV2.id ->  {
+                val reaction: FfiReaction = decodeReaction(encodedContent!!.toByteArray())
```
cameronvoell, line 182: reactionV2 in this PR mainly serves to show how proto based ContentTypes that are used in libxmtp and upstream xmtp-android/ios can be utilized in React Native. 

In this PR, we're basically converting the proto type back to json before utilizing the Reaction content in RN typescript code. An alternative would be to send the reaction content as bytes over the RN bridge and deserialize the bytes to the Reaction proto object in typescript. This pattern would generally lead to less data going over the RN bridge and would probably be a net gain for pretty much all data classes we use. However, it probably makes sense to start implementing that pattern in a separate PR.

Lastly, in a follow up, we can delete the reactionV2Codec in RN, and simply update the ContentTypeId version from 1.0 to 2.0, and the underlying SDK will switch from using JSON serializing to proto serializing without RN integrators knowing the difference. One exception is that old clients will not be able to deserialize messages sent with the proto backed reactions, but they should recognize that the version has been updated, and display the fallback instead. maintaining the two different Reaction and ReactionV2 for a short while in RN will help us test those paths.

example/patches/react-native+0.71.14.patch
```diff
@@ -0,0 +1,194 @@
+diff --git a/node_modules/react-native/sdks/hermes/hermes-engine.podspec b/node_modules/react-native/sdks/hermes/hermes-engine.podspec
```
cameronvoell, line 1: See issue here on incorrect checksum error for boost dependency. Only affects our RN example app - https://github.com/facebook/react-native/issues/42180

.changeset/config.json
```diff
@@ -1,10 +1,15 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@3.0.4/schema.json",
-  "changelog": "@changesets/cli/changelog",
+  "changelog": [
```
cameronvoell, line 3: the documentation isnt super clear on this, but hoping this will fix our changesets not creating github release / tag.

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
```diff
@@ -95,6 +102,17 @@ class ContentJson(
...
                 )
+            } else if (obj.has("reactionV2")) {
+                val reaction = obj.get("reactionV2").asJsonObject
+                return ContentJson(
+                    ContentTypeReactionV2, FfiReaction(
+                        reference = reaction.get("reference").asString,
+                        action = getReactionV2Action(reaction.get("action").asString.lowercase()),
+                        schema = getReactionV2Schema(reaction.get("schema").asString.lowercase()),
+                        content = reaction.get("content").asString,
+                        referenceInboxId = ""
```
mchenani, line 114: Question: is that intentional left empty?

mchenani (approved):

cameronvoell:
* Note, not releasing until this fix is included in the upstream SDKs: https://github.com/xmtp/libxmtp/pull/1577

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
cameronvoell: yea that property is sort of stubbed in for now, as I noticed it was included in the xmtp-js reaction type: https://github.com/xmtp/xmtp-js/blob/80aa5ec4fb24b60def222537cebcb20c818ca993/content-types/content-type-reaction/src/Reaction.ts#L24

However xmtp-react-native has an existing [ReactionContent](https://github.com/xmtp/xmtp-react-native/blob/6c593b3da9602679758279e6e02a3bf608c979fe/src/lib/types/ContentCodec.ts#L19-L24) type that didn't feel worth updating at this point since I'm not sure what `referenceInboxId` would actually be used for. 

definitely a code smell though, I added a clarifying comments here 👌: faad8de361aa2825e204bc87f04ae097973aab78

PR #609, opened 2025-02-25
Granular Control Over Creation Flow

nplasterer:
Exposes Ffi functions that allow for more granular control over client creation so signing and registering can be handled.

All of this is delicate because takes advanced understanding of the protocol to use correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "devDependencies": {
     "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
-    "@types/react": "^18.2.39",
+    "@types/react": "^19.0.0",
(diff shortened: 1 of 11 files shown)


mchenani (approved):

cameronvoell (approved):

package.json
```diff
@@ -58,7 +58,7 @@
   },
   "devDependencies": {
     "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
-    "@types/react": "^18.2.39",
+    "@types/react": "^19.0.0",
```
cameronvoell, line 61: think this should be fine for integrators since it's a dev dependency 👍

PR #615, opened 2025-03-06
Should Push for Push Notifications

nplasterer:
Adds the should_push field to messages so that push notification servers can respect them accordingly.

Should also fix rust panic on streams


mchenani (approved):

cameronvoell (approved):
🚀

PR #628, opened 2025-03-25
Allow LOCAL with a custom string

### LINKED ISSUE #626: Allow custom local urls
Like what we did here for android allow setting a custom url for local https://github.com/xmtp/xmtp-android/pull/170/files

nplasterer:
Closes https://github.com/xmtp/xmtp-react-native/issues/626

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/AuthParamsWrapper.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/AuthParamsWrapper.kt
@@ -7,6 +7,7 @@ class AuthParamsWrapper(
     val environment: String,
     val dbDirectory: String?,
     val historySyncUrl: String?,
+    val customLocalUrl: String?,
 ) {
     companion object {
         fun authParamsFromJson(authParams: String): AuthParamsWrapper {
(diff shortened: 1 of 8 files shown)


ios/XMTPModule.swift
```diff
@@ -2457,11 +2460,14 @@ public class XMTPModule: Module {
...
 
-	func createApiClient(env: String)
+	func createApiClient(env: String, customLocalUrl: String? = nil)
 		-> XMTP.ClientOptions.Api
 	{
 		switch env {
 		case "local":
+			if let customLocalUrl = customLocalUrl, !customLocalUrl.isEmpty {
+				XMTP.XMTPEnvironment.customLocalAddress = customLocalUrl
+			}
```
thierryskoda, line 2470: Maybe totally wrong here but shouldn't this be: 

```suggestion
			if let customLocalUrl = customLocalUrl, !customLocalUrl.isEmpty {
                              return XMTP.ClientOptions.Api(
	                              env: XMTP.XMTPEnvironment.custom(localAddress: customLocalUrl),
	                              isSecure: false
                              )
			}
```

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/AuthParamsWrapper.kt
```diff
@@ -7,6 +7,7 @@ class AuthParamsWrapper(
     val environment: String,
     val dbDirectory: String?,
     val historySyncUrl: String?,
+    val customLocalUrl: String?,
```
rygine, line 10: FWIW, we use `apiUrl` in the browser and node SDKs. it overrides the `env` options.

https://github.com/xmtp/xmtp-js/blob/main/sdks/node-sdk/src/Client.ts#L56-L60

nplasterer: This is in the bridge so doesn't actually get exposed or called anywhere. The way we have this setup in Android and iOS we append the http and the port. So it's not a url but the host that we allow being set to custom. Maybe in the future we can modify that but for now leaving as is.

cameronvoell (approved):

PR #648, opened 2025-04-15
Harden around forked groups

nplasterer:
This pulls in the latest libxmtp which includes two forked group fixes as well as the ability to sync a sync group.


cameronvoell (approved):

cameronvoell:
Tests looking good on ios/android

image

image

PR #765, opened 2026-01-23
deleteMessage, prepareMessage noSend, streamMessageDeletions (libxmtp 1.9.0 updates)

cameronvoell:
### Add `Client.sendSyncRequest()`, extend message preparation with `noSend`, implement `Conversations.streamMessageDeletions()`, and upgrade native SDKs to libxmtp 1.9.0 across iOS (XMTP 4.9.0) and Android (org.xmtp:android 4.9.0)
Introduce manual device sync via `Client.sendSyncRequest()`, support deferred publishing with `Conversation.prepareMessage(..., noSend)`, add real-time deletion streaming with `Conversations.streamMessageDeletions()`, implement enriched message decoding/serialization (`DecodedMessageV2`) and deletion/leave-request codecs, expose publish/delete/enriched messages APIs, and pass `appData` through group creation; bump native dependencies to XMTP 4.9.0.

#### 📍Where to Start
Start with the deletion streaming flow at `Conversations.streamMessageDeletions` in [Conversations.ts](https://github.com/xmtp/xmtp-react-native/pull/765/files#diff-c5227210ea741b70a7f1c730d3dc81ff6c90156a21cee65f6fbb29f86099c9f8), then review the platform handlers in [XMTPModule.swift](https://github.com/xmtp/xmtp-react-native/pull/765/files#diff-6260b37315249e4d7814a7eed83d9c53d15c299d753edd037d00ddd617d5a2b9) and [XMTPModule.kt](https://github.com/xmtp/xmtp-react-native/pull/765/files#diff-3d1b8496c4f066a0a1d2f776063d85d49f70545d6a94a57fe91eb6efecd50a48).

#### Changes since #765 opened

- Changed `DecodedMessageV2.from()` factory method to assign the recursively parsed reactions array or an empty array instead of passing through the decoded reactions directly [68233db]
- Updated `DecodedMessageV2.content()` and `DecodedMessageV2.decodedContent()` methods to document FFI limitations with `enrichedMessages()` for unknown or custom content types and adjusted codec handling [68233db]
- Added documentation in conversation tests explaining a removed test and the limitation where `enrichedMessages()` does not apply `JSContentCodec.decode()` transformations [68233db]
- Modified `ContentJsonV2.toJsonMap` in iOS to serialize unknown and custom content types as structured JSON [a789e15]
- Added assertions and logging to the custom content type test to validate structured JSON in `nativeContent` [a789e15]
- Consolidated `DecodedMessageUnionV2` import and removed default initialization of `deliveryStatus` in `DecodedMessageV2` [a789e15]
- Changed `MessageWrapper.encodeToObjV2` to conditionally include optional fields in result dictionary [e13af0a]
- Removed debug information upload functionality including the `uploadDebugInformation` exported function, the `XMTPDebugInformation.uploadDebugInformation` instance method, and the associated test case that verified upload key generation [df5dc25]
- Added `updateAppDataPolicy: 'allow'` field to `customPermissionsPolicySet` objects in group permissions tests [2e00272]
- Updated `reactionV2` content parsing in `ContentJson` wrapper to use `Reaction` payload type instead of `FfiReactionPayload`, changed helper functions from `getReactionV2Action` and `getReactionV2Schema` to `getReactionAction` and `getReactionSchema`, and populated `referenceInboxId` field from JSON data instead of using hardcoded empty string [cc2c241]
- Replaced `FfiReactionPayload` construction with `Reaction` construction in the reactionV2 branch of `ContentJson.fromJsonObj` parser [57d55a3]
- Changed `DeleteMessageCodec.contentType.typeId` from 'deletedMessage' to 'deleteMessage' [68d2712]
- Added test verifying native codec content type IDs match expected native SDK values [68d2712]
- Added delete message content type handling to JSON serialization [7eed895]
- Modified reaction object parsing in `ContentJson` wrapper to handle JSON null values for `referenceInboxId` [6da1e24]
- Updated package version from 5.1.0 to 5.2.0 [1f6cb9d]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 7eed895. 30 files reviewed, 51 issues evaluated, 33 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt — 0 comments posted, 5 evaluated, 4 filtered</summary>

- [line 246](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L246): The `debugEventsEnabled` parameter from `authOptions` is no longer passed to `ClientOptions`. The value is still parsed from JSON in `authParamsFromJson` and stored in `AuthParamsWrapper`, but the removal of this line means debug events configuration will be silently ignored. If `ClientOptions` still supports this parameter, users setting `debugEventsEnabled: true` will not get expected debug event behavior. <b>[ Already posted ]</b>
- [line 593](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L593): The change from `sigRequest.let` to `sigRequest?.let` in `ffiRevokeAllOtherInstallationsSignatureText` means the function will now silently return `null` if `client.ffiRevokeAllOtherInstallations()` returns null. Callers expecting a non-null signature text string may encounter unexpected null values, potentially causing `NullPointerException` downstream. <b>[ Low confidence ]</b>
- [line 2424](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L2424): Catching `Exception` on line 2424 will also catch `CancellationException`, which can interfere with structured concurrency. When the coroutine is cancelled (e.g., via `subscriptions[...].cancel()`), the `CancellationException` will be caught, logged as an error, and not rethrown. This prevents proper cancellation propagation and could cause the coroutine to not terminate cleanly. The catch block should either catch a more specific exception type or rethrow `CancellationException`. <b>[ Already posted ]</b>
- [line 2426](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/XMTPModule.kt#L2426): In the catch block at line 2426, the code calls `cancel()` on the same job that is currently executing and failing. This is effectively a no-op since the job is already completing exceptionally, and represents dead code that suggests confusion about the coroutine lifecycle. <b>[ Already posted ]</b>
</details>

<details>
<summary>android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 154](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt#L154): If `referenceInboxId` key exists in the JSON but has a `null` value (e.g., `"referenceInboxId": null`), `reaction.get("referenceInboxId")` will return `JsonNull` (not Kotlin `null`), so the safe call `?.asString` won't protect against it—calling `asString` on `JsonNull` throws `UnsupportedOperationException`. Consider using `reaction.get("referenceInboxId")?.takeIf { !it.isJsonNull }?.asString ?: ""` instead. <b>[ Already posted ]</b>
- [line 176](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt#L176): The `fromJsonObject` method does not handle `"leaveRequest"` or `"deleteMessage"` JSON keys, but `toJsonMap()` serializes them with those keys. If a caller serializes a `LeaveRequest` or `DeleteMessageRequest` via `toJsonMap()` and then attempts to deserialize it via `fromJson()`, it will throw `"Unknown content type"` exception instead of round-tripping correctly. <b>[ Already posted ]</b>
</details>

<details>
<summary>android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt — 0 comments posted, 10 evaluated, 7 filtered</summary>

- [line 143](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L143): In `fromJsonObject` at line 143, the `reactionV2` branch creates a `ContentJson` with `FfiReactionPayload` as content type, but `ContentJson` (shown in the diff) was updated to use the SDK `Reaction` type instead. The comment on line 147 in `ContentJson` explicitly states "ReactionV2Codec encodes Reaction" - using `FfiReactionPayload` here will cause a type mismatch when the codec tries to encode the content, likely resulting in a `ClassCastException` or encoding failure. <b>[ Already posted ]</b>
- [line 143](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L143): In `ContentJsonV2.fromJsonObject`, when parsing `reactionV2`, the code still creates `FfiReactionPayload` (line 143) and uses `getReactionV2Action`/`getReactionV2Schema` (lines 145-146), but the `ReactionV2Codec` was updated to encode `Reaction` type instead. The diff comment in `ContentJson.kt` explicitly states "Use SDK Reaction type (not FfiReactionPayload); ReactionV2Codec encodes Reaction." This inconsistency means messages created via `ContentJsonV2.fromJson` with `reactionV2` content will fail when encoded because the codec expects `Reaction` but receives `FfiReactionPayload`. <b>[ Already posted ]</b>
- [line 143](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L143): In `ContentJsonV2.fromJsonObject()`, the `reactionV2` handling creates a `ContentJson` with `FfiReactionPayload` type at line 143, but the diff shows that `ContentJson.kt` was updated to use `Reaction` type instead. The `ReactionV2Codec` (per the comment in the updated `ContentJson.kt`) encodes `Reaction`, not `FfiReactionPayload`. This type mismatch will cause encoding failures or incorrect behavior when processing reactionV2 content through `ContentJsonV2`. <b>[ Already posted ]</b>
- [line 143](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L143): In `fromJsonObject`, the `reactionV2` branch creates a `ContentJson` with `FfiReactionPayload` as the content type (lines 142-151), but according to the changes in `ContentJson.kt`, `ReactionV2Codec` now expects the SDK `Reaction` type instead. This is an incomplete change - `ContentJson.kt` was updated to use `Reaction` with `getReactionAction`/`getReactionSchema`, but `ContentJsonV2.kt` still uses the old `FfiReactionPayload` with `getReactionV2Action`/`getReactionV2Schema`. This type mismatch will cause encoding failures when sending reactionV2 messages via `ContentJsonV2.fromJsonObject`. <b>[ Already posted ]</b>
- [line 145](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L145): At line 145-146, `getReactionV2Action` and `getReactionV2Schema` are used which return `FfiReactionAction` and `FfiReactionSchema` types, but since the code constructs `FfiReactionPayload` (which requires these FFI types), if this is corrected to use `Reaction` type instead, these functions should be replaced with `getReactionAction` and `getReactionSchema` which return the SDK-compatible types. <b>[ Already posted ]</b>
- [line 145](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L145): In `ContentJsonV2.fromJsonObject()`, the `reactionV2` branch uses `getReactionV2Action()` and `getReactionV2Schema()` at lines 145-146, but `ContentJson.kt` was updated to use `getReactionAction()` and `getReactionSchema()` instead. These functions return different types (`FfiReactionAction`/`FfiReactionSchema` vs SDK `ReactionAction`/`ReactionSchema`), causing type incompatibility with the expected `Reaction` constructor parameters. <b>[ Already posted ]</b>
- [line 321](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJsonV2.kt#L321): At line 321, there is a redundant safe cast `(content as? DeleteMessageRequest)?.messageId` when `content` is already declared as `DeleteMessageRequest?` on line 318. This is unnecessary but harmless. <b>[ Low confidence ]</b>
</details>

<details>
<summary>android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/EnrichedMessageQueryParamsWrapper.kt — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 32](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/EnrichedMessageQueryParamsWrapper.kt#L32): If `paramsJson` contains malformed JSON that is not an object (e.g., `"hello"`, `123`, or `[]`), the `.asJsonObject` call on line 32 will throw an `IllegalStateException`. Consider wrapping this in a try-catch or validating the JSON structure. <b>[ Already posted ]</b>
- [line 36](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/EnrichedMessageQueryParamsWrapper.kt#L36): When a JSON field exists but has a `null` value (e.g., `{"limit": null}`), `jsonOptions.has("limit")` returns `true`, but calling `.asInt`, `.asLong`, or `.asString` on a `JsonNull` element will throw an `UnsupportedOperationException`. This affects all field extractions (lines 36, 43, 50, 57, 72, 79, 86, 93). The code should check `!jsonOptions.get("field").isJsonNull` before calling the type conversion methods. <b>[ Already posted ]</b>
- [line 64](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/EnrichedMessageQueryParamsWrapper.kt#L64): If `excludeSenderInboxIds` exists in the JSON but is not an array (e.g., a string or object), `getAsJsonArray("excludeSenderInboxIds")` on line 64 will throw an `IllegalStateException`. <b>[ Already posted ]</b>
- [line 65](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/EnrichedMessageQueryParamsWrapper.kt#L65): If the `excludeSenderInboxIds` array contains `null` elements in the JSON (e.g., `{"excludeSenderInboxIds": ["id1", null]}`), the `it.asString` call on line 65 will throw an `UnsupportedOperationException` when it encounters the null element. <b>[ Already posted ]</b>
</details>

<details>
<summary>android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/PermissionPolicySetWrapper.kt — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 55](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/PermissionPolicySetWrapper.kt#L55): In `createPermissionPolicySetFromJson`, calling `jsonObj.get("updateAppDataPolicy").asString` will throw a `NullPointerException` if the JSON input doesn't contain the `updateAppDataPolicy` field. This breaks backward compatibility with any existing JSON data that was created before this field was added. The `get()` method returns `null` for missing keys, and calling `.asString` on `null` crashes. <b>[ Already posted ]</b>
</details>

<details>
<summary>ios/Wrappers/CreateGroupParamsWrapper.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 22](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/CreateGroupParamsWrapper.swift#L22): The cast `as? Int64` for `disappearStartingAtNs` and `retentionDurationInNs` will silently fail even when valid numeric values are present in the JSON. `JSONSerialization` returns `NSNumber` for numeric values, which doesn't bridge directly to `Int64`. This will cause `DisappearingMessageSettings` to never be created even when the values are provided. Use `(jsonOptions["disappearStartingAtNs"] as? NSNumber)?.int64Value` instead. <b>[ Already posted ]</b>
</details>

<details>
<summary>ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 36](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift#L36): The casts `as? Int64` for `beforeNs`, `afterNs`, `insertedAfterNs`, and `insertedBeforeNs` will silently fail and return `nil` even when valid numeric values are present in the JSON. `JSONSerialization` returns `NSNumber` for numeric values, which doesn't bridge directly to `Int64`. Use `(jsonOptions["beforeNs"] as? NSNumber)?.int64Value` pattern instead. <b>[ Already posted ]</b>
</details>

<details>
<summary>ios/Wrappers/MessageWrapper.swift — 1 comment posted, 6 evaluated, 4 filtered</summary>

- [line 31](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/MessageWrapper.swift#L31): The method `encodeToObjV2` is marked with `throws` but never actually throws an error - all internal throwing calls use `try?` to swallow errors. While not a runtime crash, this is misleading since `encodeV2` uses `try encodeToObjV2(model)` expecting potential errors that will never come from this method. <b>[ Low confidence ]</b>
- [line 33](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/MessageWrapper.swift#L33): Using `try?` with `compactMap` on line 33-34 silently drops any reactions that fail to encode. If `encodeToObjV2(reaction)` throws for any reaction, it will be filtered out of the results without any error or warning. This causes silent data loss where the consumer receives an incomplete `reactions` array and a `reactionCount` that doesn't match the number of reactions returned. <b>[ Already posted ]</b>
- [line 123](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/MessageWrapper.swift#L123): Similar to the action issue, changing from `ReactionV2Schema.fromString()` to `ReactionSchema(rawValue:)` may fail silently. `ReactionV2Schema.fromString` handles "unicode", "shortcode", "custom" with a fallback to `.unknown`, but `ReactionSchema(rawValue:)` returns `nil` for unrecognized values. If the raw value strings differ between the V2 helper and `ReactionSchema` enum, the schema would be `nil`. <b>[ Already posted ]</b>
- [line 385](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/ios/Wrappers/MessageWrapper.swift#L385): Inconsistent `scheme` value between `ContentTypeRemoteAttachment` (line 385: `"https://"`) and `ContentTypeMultiRemoteAttachment` (line 395: `"https"`). If downstream consumers expect a consistent format, this mismatch will cause parsing or URL reconstruction failures. <b>[ Already posted ]</b>
</details>

<details>
<summary>src/lib/Conversations.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 591](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/Conversations.ts#L591): The `streamMessageDeletions` method assigns the new subscription to `this.subscriptions[EventTypes.MessageDeletion]` without removing any existing subscription at that key. If `streamMessageDeletions` is called multiple times on the same `Conversations` instance (e.g., in a component lifecycle or multiple subscribers), the reference to the previous subscription is overwritten and lost. Since `cancelStreamMessageDeletions` uses this key to find the subscription to remove, the overwritten subscription can never be cleaned up via that method. This results in a permanent memory leak of the event listener and duplicate execution of callbacks for incoming deletion events. <b>[ Already posted ]</b>
</details>

<details>
<summary>src/lib/DecodedMessageV2.ts — 0 comments posted, 7 evaluated, 5 filtered</summary>

- [line 51](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/DecodedMessageV2.ts#L51): The `DecodedMessageV2.from` method incorrectly propagates the parent message's generic `ContentType` to nested `reactions`. When a parent message (e.g., `TextCodec`) is parsed, its reactions are instantiated as `DecodedMessageV2<TextCodec>` via `fromObject<ContentType>`. However, at runtime, `reaction.content()` decodes the payload using the reaction's actual codec (found via `this.contentTypeId` from the JSON), returning a reaction object. TypeScript expects a string (due to the `TextCodec` generic), leading to runtime crashes if string methods (like `toUpperCase`) are called on the returned object. <b>[ Already posted ]</b>
- [line 60](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/DecodedMessageV2.ts#L60): The `DecodedMessageV2.from` method deserializes JSON directly into the class properties without converting date strings to `Date` objects. `JSON.parse` returns strings for date fields (like `sentAt` and `expiresAt`), but the class properties are typed as `Date`. When a consumer attempts to call `Date` methods (e.g., `msg.sentAt.getTime()`) on these properties, the application will crash with `TypeError: ... is not a function` because the runtime value is a string. <b>[ Already posted ]</b>
- [line 149](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/DecodedMessageV2.ts#L149): The `DecodedMessageV2` class fails to validate the existence of `nativeContent` on the parsed JSON object before assigning it. If the native layer returns a partial or malformed JSON object where `nativeContent` is missing (undefined), the `content()` method will throw a `TypeError` (e.g., `Cannot read properties of undefined (reading 'encoded')`) when accessing `this.nativeContent.encoded`. <b>[ Already posted ]</b>
- [line 202](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/DecodedMessageV2.ts#L202): The logic in `DecodedMessageV2.content()` introduces a critical flaw in how decoders are selected for native content. The loop at lines 199-210 iterates over all registered codecs. The condition `allowEmptyProperties.some(...)` (lines 202-204) checks if `nativeContent` has *any* allowed empty property (e.g., `'text'`) but fails to verify if that property corresponds to the *current* `codec` in the loop. <b>[ Already posted ]</b>
- [line 208](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/DecodedMessageV2.ts#L208): In `DecodedMessageV2.content()`, the logic iterates over all registered codecs in `Client.codecRegistry` to find a matching `NativeContentCodec`. However, it accesses `codec.contentKey` without first verifying that the property exists on the codec object. `Client.codecRegistry` can contain both `JSContentCodec` (which does not have `contentKey`) and `NativeContentCodec` (which does). Accessing `codec.contentKey` on a `JSContentCodec` will return `undefined`. While `this.nativeContent[undefined]` is generally safe in JS (returns undefined), if `allowEmptyProperties` logic is not met, the loop continues. If a `JSContentCodec` is encountered, the code does not crash immediately at the property check, but if the loop relies on `codec` being a `NativeContentCodec` for the `decode` call later, we must ensure it is one. More critically, the check `('contentKey' in codec ...)` is present, but TypeScript runtime behavior for `in` operator on objects is safe. The issue here is subtle: if `this.nativeContent.hasOwnProperty(prop)` is true for an empty property (like `text`), the code executes `(codec as NativeContentCodec<...>).decode(this.nativeContent)`. If the current `codec` in the loop happens to be a `JSContentCodec` (which lacks `decode(NativeMessageContent)` signature or implementation compatible with it, usually expecting `EncodedContent`), this will crash or throw a type error at runtime because `JSContentCodec.decode` takes `EncodedContent`, not `NativeMessageContent`. The loop iterates `Object.values(Client.codecRegistry)`, so it mixes both types. <b>[ Already posted ]</b>
</details>

<details>
<summary>src/lib/Dm.ts — 0 comments posted, 4 evaluated, 3 filtered</summary>

- [line 143](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/Dm.ts#L143): The `prepareMessage` method accepts `SendOptions` (which includes `shouldPush`), but fails to pass this parameter to the underlying native calls. Specifically, in `_prepareWithJSCodec` (line 171) and the direct `XMTP.prepareMessage` call (line 159), the `shouldPush` option is dropped. This means users cannot control push notification behavior for prepared messages (e.g., suppressing push for silent messages), unlike the `send` method which correctly processes `opts.shouldPush`. <b>[ Already posted ]</b>
- [line 162](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/Dm.ts#L162): The `prepareMessage` method accepts `opts: SendOptions` (which includes `shouldPush?: boolean`), but the implementation discards the `shouldPush` property in both execution paths. When calling `_prepareWithJSCodec` (line 162), `shouldPush` is not passed, and `_prepareWithJSCodec` itself does not accept it as an argument. Similarly, in the direct native call path (line 170), `XMTP.prepareMessage` is called without any push configuration. Because `prepareMessage` performs an optimistic send by default (when `noSend` is false), users invoking `prepareMessage` with `shouldPush: false` will have this preference ignored, resulting in unintended push notifications if the underlying default is to push. <b>[ Already posted ]</b>
- [line 199](https://github.com/xmtp/xmtp-react-native/blob/7eed89508fedbfd287fe0ec95b53597bb8c6f70c/src/lib/Dm.ts#L199): The `publishMessage` method triggers the sending of a previously prepared message, but because `prepareMessage` dropped the `shouldPush` preference during the initial storage phase, `publishMessage` has no way to retrieve or respect the user's original intent regarding push notifications. This results in the message being published with default push behavior (typically enabled) regardless of the options provided during preparation. <b>[ Already posted ]</b>
</details>

</details>

### DIFF
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
@@ -144,14 +144,14 @@ class ContentJson(
                 )
             } else if (obj.has("reactionV2")) {
                 val reaction = obj.get("reactionV2").asJsonObject
+                // Use SDK Reaction type (not FfiReactionPayload); ReactionV2Codec encodes Reaction.
                 return ContentJson(
-                    ContentTypeReactionV2, FfiReactionPayload(
+                    ContentTypeReactionV2, Reaction(
                         reference = reaction.get("reference").asString,
-                        action = getReactionV2Action(reaction.get("action").asString.lowercase()),
-                        schema = getReactionV2Schema(reaction.get("schema").asString.lowercase()),
+                        action = getReactionAction(reaction.get("action").asString.lowercase()),
+                        schema = getReactionSchema(reaction.get("schema").asString.lowercase()),
                         content = reaction.get("content").asString,
-                        // Update if we add referenceInboxId to ../src/lib/types/ContentCodec.ts#L19-L24
-                        referenceInboxId = ""
+                        referenceInboxId = reaction.get("referenceInboxId")?.asString ?: ""
                     )
                 )
             } else if (obj.has("reply")) {
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
@@ -27,7 +27,10 @@ import org.xmtp.android.library.codecs.MultiRemoteAttachment
 import org.xmtp.android.library.codecs.Reaction
 import org.xmtp.android.library.codecs.ReactionCodec
 import org.xmtp.android.library.codecs.ReactionV2Codec
+import org.xmtp.android.library.codecs.ContentTypeDeleteMessageRequest
 import org.xmtp.android.library.codecs.ContentTypeLeaveRequest
+import org.xmtp.android.library.codecs.DeleteMessageCodec
+import org.xmtp.android.library.codecs.DeleteMessageRequest
 import org.xmtp.android.library.codecs.LeaveRequest
 import org.xmtp.android.library.codecs.LeaveRequestCodec
 import org.xmtp.android.library.codecs.ReadReceipt
@@ -75,7 +78,7 @@ class ContentJson(
             Client.register(GroupUpdatedCodec())
             Client.register(ReactionV2Codec())
             Client.register(LeaveRequestCodec())
-//            Client.register(DeleteMessageCodec())
+            Client.register(DeleteMessageCodec())
         }
 
         fun fromJsonObject(obj: JsonObject): ContentJson {
@@ -293,11 +296,11 @@ class ContentJson(
                 )
             )
 
-//            ContentTypeDeleteMessage.id -> mapOf(
-//                "deleteMessage" to mapOf(
-//                    "messageId" to ((content as? DeleteMessage)?.messageId ?: "")
-//                )
-//            )
+           ContentTypeDeleteMessageRequest.id -> mapOf(
+               "deleteMessage" to mapOf(
+                   "messageId" to ((content as? DeleteMessageRequest)?.messageId ?: "")
+               )
+           )
 
             else -> {
                 val json = JsonObject()
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
@@ -27,6 +27,9 @@ import org.xmtp.android.library.codecs.MultiRemoteAttachment
 import org.xmtp.android.library.codecs.Reaction
 import org.xmtp.android.library.codecs.ReactionCodec
 import org.xmtp.android.library.codecs.ReactionV2Codec
+import org.xmtp.android.library.codecs.ContentTypeLeaveRequest
+import org.xmtp.android.library.codecs.LeaveRequest
+import org.xmtp.android.library.codecs.LeaveRequestCodec
 import org.xmtp.android.library.codecs.ReadReceipt
 import org.xmtp.android.library.codecs.ReadReceiptCodec
 import org.xmtp.android.library.codecs.RemoteAttachment
@@ -71,6 +74,8 @@ class ContentJson(
             Client.register(ReadReceiptCodec())
             Client.register(GroupUpdatedCodec())
             Client.register(ReactionV2Codec())
+            Client.register(LeaveRequestCodec())
+            Client.register(DeleteMessageCodec())
         }
 
         fun fromJsonObject(obj: JsonObject): ContentJson {
@@ -279,6 +284,18 @@ class ContentJson(
                             "fieldName" to it.fieldName,
                         )
                     },
+                )
+            )
+
+            ContentTypeLeaveRequest.id -> mapOf(
+                "leaveRequest" to mapOf(
+                    "authenticatedNote" to ((content as? LeaveRequest)?.authenticatedNote ?: "")
+                )
+            )
+
+            ContentTypeDeleteMessage.id -> mapOf(
+                "deleteMessage" to mapOf(
+                    "messageId" to ((content as? DeleteMessage)?.messageId ?: "")
                 )
             )
 
--- a/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
+++ b/android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
@@ -151,7 +151,7 @@ class ContentJson(
                         action = getReactionAction(reaction.get("action").asString.lowercase()),
                         schema = getReactionSchema(reaction.get("schema").asString.lowercase()),
                         content = reaction.get("content").asString,
-                        referenceInboxId = reaction.get("referenceInboxId")?.asString ?: ""
+                        referenceInboxId = reaction.get("referenceInboxId")?.takeIf { !it.isJsonNull }?.asString ?: ""
                     )
                 )
             } else if (obj.has("reply")) {
--- a/ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift
+++ b/ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift
@@ -1,0 +1,57 @@
+import Foundation
+import XMTP
+
+struct EnrichedMessageQueryParamsWrapper {
+	let limit: Int?
+	let beforeNs: Int64?
+	let afterNs: Int64?
+	let direction: String?
+	let excludeSenderInboxIds: [String]?
+	let deliveryStatus: String?
+	let insertedAfterNs: Int64?
+	let insertedBeforeNs: Int64?
+	let sortBy: String?
+
+	static func fromJson(_ paramsJson: String) -> EnrichedMessageQueryParamsWrapper {
+		guard !paramsJson.isEmpty else {
+			return EnrichedMessageQueryParamsWrapper(
+				limit: nil,
+				beforeNs: nil,
+				afterNs: nil,
+				direction: nil,
+				excludeSenderInboxIds: nil,
+				deliveryStatus: nil,
+				insertedAfterNs: nil,
+				insertedBeforeNs: nil,
+				sortBy: nil
+			)
+		}
+
+		let data = paramsJson.data(using: .utf8) ?? Data()
+		let jsonOptions =
+			(try? JSONSerialization.jsonObject(with: data, options: []))
+			as? [String: Any] ?? [:]
+
+		let limit = jsonOptions["limit"] as? Int
+		let beforeNs = jsonOptions["beforeNs"] as? Int64
+		let afterNs = jsonOptions["afterNs"] as? Int64
+		let direction = jsonOptions["direction"] as? String
+		let excludeSenderInboxIds = jsonOptions["excludeSenderInboxIds"] as? [String]
+		let deliveryStatus = jsonOptions["deliveryStatus"] as? String
+		let insertedAfterNs = jsonOptions["insertedAfterNs"] as? Int64
+		let insertedBeforeNs = jsonOptions["insertedBeforeNs"] as? Int64
+		let sortBy = jsonOptions["sortBy"] as? String
+
+		return EnrichedMessageQueryParamsWrapper(
+			limit: limit,
+			beforeNs: beforeNs,
+			afterNs: afterNs,
+			direction: direction,
+			excludeSenderInboxIds: excludeSenderInboxIds,
+			deliveryStatus: deliveryStatus,
+			insertedAfterNs: insertedAfterNs,
+			insertedBeforeNs: insertedBeforeNs,
+			sortBy: sortBy
+		)
+	}
+}
--- a/ios/Wrappers/MessageWrapper.swift
+++ b/ios/Wrappers/MessageWrapper.swift
@@ -115,13 +115,12 @@ struct ContentJson {
 				schema: ReactionSchema(rawValue: reaction["schema"] as? String ?? "")
 			))
 		} else if let reaction = obj["reactionV2"] as? [String: Any] {
-            return ContentJson(type: ContentTypeReactionV2, content: FfiReactionPayload(
+			// Use SDK Reaction type (not FfiReactionPayload); ReactionV2Codec encodes Reaction.
+			return ContentJson(type: ContentTypeReactionV2, content: Reaction(
 				reference: reaction["reference"] as? String ?? "",
-				// Update if we add referenceInboxId to ../src/lib/types/ContentCodec.ts#L19-L24
-                referenceInboxId: "",
-				action: ReactionV2Action.fromString(reaction["action"] as? String ?? ""),
+				action: ReactionAction(rawValue: reaction["action"] as? String ?? ""),
 				content: reaction["content"] as? String ?? "",
-				schema: ReactionV2Schema.fromString(reaction["schema"] as? String ?? "")
+				schema: ReactionSchema(rawValue: reaction["schema"] as? String ?? "")
 			))
 		}else if let reply = obj["reply"] as? [String: Any] {
 			guard let nestedContent = reply["content"] as? [String: Any] else {
--- a/ios/Wrappers/MessageWrapper.swift
+++ b/ios/Wrappers/MessageWrapper.swift
@@ -300,11 +300,11 @@ struct ContentJson {
 			return ["leaveRequest": [
 				"authenticatedNote": noteString
 			]]
-//		case ContentTypeDeleteMessage.id where content is XMTP.DeleteMessage:
-//			let deleteMessage = content as! XMTP.DeleteMessage
-//			return ["deleteMessage": [
-//				"messageId": deleteMessage.messageId,
-//			]]
+		case ContentTypeDeleteMessageRequest.id where content is XMTP.DeleteMessageRequest:
+			let deleteMessage = content as! XMTP.DeleteMessageRequest
+			return ["deleteMessage": [
+				"messageId": deleteMessage.messageId,
+			]]
 		default:
 			if let encodedContent, let encodedContentJSON = try? encodedContent.jsonString() {
 				return ["encoded": encodedContentJSON]
--- a/ios/Wrappers/MessageWrapper.swift
+++ b/ios/Wrappers/MessageWrapper.swift
@@ -34,7 +34,7 @@ struct MessageWrapper {
 			try? encodeToObjV2(reaction)
 		}
 		let reactionCount = reactionsList.count
-		return [
+		var result: [String: Any] = [
 			"id": model.id,
 			"conversationId": model.conversationId,
 			"contentTypeId": model.contentTypeId.description,
@@ -43,14 +43,21 @@ struct MessageWrapper {
 			"sentAt": model.sentAtNs / 1_000_000,
 			"sentAtNs": model.sentAtNs,
 			"insertedAtNs": model.insertedAtNs,
-			"expiresAtNs": model.expiresAtNs as Any,
-			"expiresAt": model.expiresAt.map { Int64($0.timeIntervalSince1970 * 1000) } as Any,
-			"fallbackText": (try? model.fallback) as Any,
 			"deliveryStatus": model.deliveryStatus.rawValue.uppercased(),
 			"reactions": reactions,
 			"hasReactions": reactionCount > 0,
 			"reactionCount": Int64(reactionCount)
 		]
+		if let expiresAtNs = model.expiresAtNs {
+			result["expiresAtNs"] = expiresAtNs
+		}
+		if let expiresAt = model.expiresAt {
+			result["expiresAt"] = Int64(expiresAt.timeIntervalSince1970 * 1000)
+		}
+		if let fallbackText = try? model.fallback {
+			result["fallbackText"] = fallbackText
+		}
+		return result
 	}
 
 	static func encodeV2(_ model: XMTP.DecodedMessageV2) throws -> String {
--- a/ios/Wrappers/MessageWrapper.swift
+++ b/ios/Wrappers/MessageWrapper.swift
@@ -437,16 +437,61 @@ struct ContentJsonV2 {
 			
 		default:
 			// For unknown/custom content types, try to serialize the content
+			// Match Android's approach: check types first, then use fallback
 			if let content: Any = try? message.content() {
-				// Try to convert to JSON-serializable format
-				let contentString = String(describing: content)
+				// First check if content is already a dictionary (most common case)
+				if let dict = content as? [String: Any] {
+					return ["unknown": [
+						"contentTypeId": contentTypeId.description,
+						"content": dict
+					]]
+				}
+				// Check if it's an array
+				if let array = content as? [Any] {
+					return ["unknown": [
+						"contentTypeId": contentTypeId.description,
+						"content": array
+					]]
+				}
+				// Check if it's a simple value (string, number, bool)
+				if let str = content as? String {
+					return ["unknown": [
+						"contentTypeId": contentTypeId.description,
+						"content": str
+					]]
+				}
+				if let num = content as? NSNumber {
+					return ["unknown": [
+						"contentTypeId": contentTypeId.description,
+						"content": num
+					]]
+				}
+				// For custom types, check if JSONSerialization can handle it
+				if JSONSerialization.isValidJSONObject(content) {
+					do {
+						let jsonData = try JSONSerialization.data(withJSONObject: content, options: [])
+						if let jsonObject = try JSONSerialization.jsonObject(with: jsonData) as? [String: Any] {
+							return ["unknown": [
+								"contentTypeId": contentTypeId.description,
+								"content": jsonObject
+							]]
+						}
+					} catch {
+						// Fall through to fallback
+					}
+				}
+				// Fall back to fallback text for non-serializable types
+				let fallbackText = (try? message.fallback) ?? "Unsupported content type"
 				return ["unknown": [
 					"contentTypeId": contentTypeId.description,
-					"content": contentString
+					"fallback": fallbackText
 				]]
 			}
+			// Content decoding returned nil
+			let fallbackText = (try? message.fallback) ?? "Failed to decode content"
 			return ["unknown": [
-				"contentTypeId": contentTypeId.description
+				"contentTypeId": contentTypeId.description,
+				"fallback": fallbackText
 			]]
 		}
 	}
--- a/ios/Wrappers/MessageWrapper.swift
+++ b/ios/Wrappers/MessageWrapper.swift
@@ -27,6 +27,36 @@ struct MessageWrapper {
 		let obj = try encodeToObj(model)
 		return try obj.toJson()
 	}
+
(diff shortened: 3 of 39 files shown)


ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import XMTP
+
+struct EnrichedMessageQueryParamsWrapper {
```
neekolas, line 4: Aren't the query params the same between enriched messages and regular messages?

src/lib/DecodedMessageV2.ts
```diff
@@ -0,0 +1,209 @@
...
+> {
+  id: MessageId
+  conversationId: ConversationId
+  senderInboxId: InboxId
+  sentAt: Date
+  sentAtNs: number // timestamp in nanoseconds
+  insertedAtNs: number
+  expiresAtNs: number | undefined
+  expiresAt: Date | undefined
+  deliveryStatus: MessageDeliveryStatus = MessageDeliveryStatus.PUBLISHED
```
neekolas: Do we need a default for this field? Shouldn't every message have a delivery status?

src/lib/DecodedMessageV2.ts
```diff
@@ -0,0 +1,209 @@
...
+        ExtractDecodedType<ContentType>
+      >
+      if (!codec) {
+        throw new Error(
+          `no content type found for unknown content: ${JSON.stringify(this.nativeContent)}`
+        )
+      }
+      // The content is already JSON-stringified, parse it directly
+      if (unknownContent.content) {
+        return JSON.parse(
```
neekolas, line 191: Shouldn't we be trying to use the codec on this, rather than just returning the JSON parsed content?

cameronvoell: think there is a limitation with using custom content types defined using the existing JSContentCodec interface in RN. that interface expects you to pass in EncodedContent, but the native sdks only make the decoded content available for custom content types on the DecodedContentV2 Object. 

I think how it's handled now will work for most custom content types, so maybe we just flag it as a limitation in RN for enrichedMessages, and try and fix this in next release. wdyt? @neekolas

src/lib/DecodedMessageV2.ts
cameronvoell: removed 👍

ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift
cameronvoell: RN was not caught up to android ios on available query options for `messages()`. im not sure if messages is the only function that is missing options that ios and android have. we could update these as well, but I think my goal was to initially bound the updates in this release to supporting `enrichedMessages` and `DecodedMessagesV2`.

How about in subsequent release along with implementing style guide updates we do an audit of other parameters / query options that were added to android / ios that are missing on react native?

ios/Wrappers/MessageWrapper.swift
```diff
@@ -27,6 +27,36 @@ struct MessageWrapper {
...
+		return [
+			"id": model.id,
+			"conversationId": model.conversationId,
+			"contentTypeId": model.contentTypeId.description,
+			"nativeContent": ContentJsonV2.toJsonMap(model),
+			"senderInboxId": model.senderInboxId,
+			"sentAt": model.sentAtNs / 1_000_000,
+			"sentAtNs": model.sentAtNs,
+			"insertedAtNs": model.insertedAtNs,
+			"expiresAtNs": model.expiresAtNs as Any,
```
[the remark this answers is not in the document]
cameronvoell: fixed e13af0a8ba6ff979fe3314744bdb6d1f384ac724

ios/Wrappers/EnrichedMessageQueryParamsWrapper.swift
neekolas: Fair enough. Fine as long as we eventually reconcile the two

src/lib/DecodedMessageV2.ts
neekolas: OK. Maybe make a ticket so we don't forget. 

Ideally we get everything back to the flow of everything being either:

1. We can decode it in Rust
2. We can decode it using a custom codec provided by the developer 
3. We can't decode it

neekolas (approved):

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
```diff
@@ -282,6 +290,18 @@ class ContentJson(
                 )
             )
 
+            ContentTypeLeaveRequest.id -> mapOf(
```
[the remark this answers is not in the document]
cameronvoell: this should be fine since we recommend sendLeaveRequest and deleteMessage functions

android/src/main/java/expo/modules/xmtpreactnativesdk/wrappers/ContentJson.kt
```diff
@@ -136,14 +144,14 @@ class ContentJson(
...
+                    ContentTypeReactionV2, Reaction(
                         reference = reaction.get("reference").asString,
-                        action = getReactionV2Action(reaction.get("action").asString.lowercase()),
-                        schema = getReactionV2Schema(reaction.get("schema").asString.lowercase()),
+                        action = getReactionAction(reaction.get("action").asString.lowercase()),
+                        schema = getReactionSchema(reaction.get("schema").asString.lowercase()),
                         content = reaction.get("content").asString,
-                        // Update if we add referenceInboxId to ../src/lib/types/ContentCodec.ts#L19-L24
-                        referenceInboxId = ""
+                        referenceInboxId = reaction.get("referenceInboxId")?.asString ?: ""
```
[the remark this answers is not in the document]
cameronvoell: referenceInboxId shouldnt ever be null, but we can add. this to be defensive

src/lib/DecodedMessageV2.ts
cameronvoell: issue here: https://github.com/xmtp/xmtp-react-native/issues/766

